PR #14, opened 2022-05-11
refactor: monorepo preparations for a worker package

spalt08:
Created the `shared` package that is going to be used for `public-api`, `worker` and `web` (API types).


packages/shared/src/internalApi/types.ts
```diff
@@ -0,0 +1,18 @@
+export namespace Internal {
```
ctizen, line 1: I'd suggest not to use namespaces at all, especially with generic names like this

zardak: I'd opt in for namespace here, with a more specific name like `GradejsInternal` or `InternalApi`

spalt08: fixed in 274dcf8

zardak (approved):

PR #18, opened 2022-05-18
ci: implement github action tests

spalt08:


packages/web/package.json
```diff
@@ -31,8 +32,8 @@
     "@types/react": "^17.0.21",
     "@types/react-dom": "^17.0.9",
     "@types/react-gtm-module": "^2.0.1",
-    "@typescript-eslint/eslint-plugin": "^4.31.1",
-    "@typescript-eslint/parser": "^4.31.1",
+    "@typescript-eslint/eslint-plugin": "^5.25.0",
```
ctizen, line 35: I guess we should copy here the .eslintrc.js from the private repo. Probably it will result in numeric errors, maybe create an issue for this?

ctizen (approved):

zardak (approved):

PR #24, opened 2022-05-24
feat: Detected package version range and estimated size

zardak:
This brings back approximate package size and adds distinct fields for version set and version range.

### DIFF
--- a/packages/web/src/components/ui/Package/Package.tsx
+++ b/packages/web/src/components/ui/Package/Package.tsx
@@ -3,28 +3,34 @@ import clsx from 'clsx';
 import React from 'react';
 import styles from './Package.module.scss';
 
+export type DetectedPackageData = {
+  packageName: string;
+  possiblePackageVersions: string[];
+  packageVersionRange: string;
+  packageMetadata?: {
+    approximateByteSize?: number;
+  };
+}
+
 export type Props = {
   className?: string;
   variant?: 'grid' | 'lines';
-  pkg: string;
-  // size?: number;
-  // bundleSize?: number;
+  pkg: DetectedPackageData;
 };
 
 export default function Package({ className, variant = 'grid', pkg }: Props) {
-  const idx = pkg.lastIndexOf('@');
-  const name = pkg.slice(0, idx);
-  const version = pkg.slice(idx + 1);
 
   return (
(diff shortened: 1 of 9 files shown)


packages/web/src/components/ui/Package/Package.tsx
```diff
@@ -3,28 +3,34 @@ import clsx from 'clsx';
...
-        <div className={styles.name}>{name}</div>
-        <div className={styles.version}>{version}</div>
+        <div className={styles.name}>{pkg.packageName}</div>
+        <div className={styles.version}>{pkg.packageVersionRange}</div>
       </div>
       <div className={styles.meta}>
-        {/* <span className={styles.size}>23B</span>
-        <span className={styles.percent}>23%</span> */}
+        {!!pkg.packageMetadata?.approximateByteSize && (
+          <span className={styles.size}>pkg.packageMetadata.approximateByteSizeB</span>
```
spalt08, line 31: There's a mistake probably:
```{pkg.packageMetadata.approximateByteSize}```

**Review by spalt08 — COMMENTED** [review]
Please also consider removing the following lines from the `requestWebPageParse` function: 
```
// Clear cached results
getRepository(WebPagePackage).delete({ hostname: getHostnameFromUrl(url) }),
```

We needed them for clearing the cache, indexed by only `hostname` field. It caused some caching issues, like package duplication: `react@*` and `react@17.0.2` during the upsert query inside `syncWebPage`.

Now we have both `hostname` and `packageName` index keys, so we don't need to clear the whole cache anymore.

spalt08 (approved):

zardak:
> Please also consider removing the following lines from the `requestWebPageParse` function:
> 
> ```
> // Clear cached results
> getRepository(WebPagePackage).delete({ hostname: getHostnameFromUrl(url) }),
> ```
> 
> We needed them for clearing the cache, indexed by only `hostname` field. It caused some caching issues, like package duplication: `react@*` and `react@17.0.2` during the upsert query inside `syncWebPage`.
> 
> Now we have both `hostname` and `packageName` index keys, so we don't need to clear the whole cache anymore.

I'd rather not mix-and-match previous results, we definitely don't want to show packages that are no longer in the bundle.

PR #26, opened 2022-05-27
feat: extended registry metadata response

spalt08:
- More useful fields for the package metadata entity
- Updated process for the package sync
- Skip flaky tests with the NPM registry

### DIFF
--- a/packages/worker/src/npmRegistry/api.ts
+++ b/packages/worker/src/npmRegistry/api.ts
@@ -103,7 +103,7 @@ function isValidUrl(maybeUrl: string) {
   return url.protocol === 'http:' || url.protocol === 'https:';
 }
 
-function getRepositoryUrl(repository?: string) {
+export function getRepositoryUrl(repository?: string) {
   let url;
 
   if (!repository) {
@@ -111,13 +111,13 @@ function getRepositoryUrl(repository?: string) {
   }
 
   // (git+)https://github.com/facebook/react(.git)
-  if ((url = repository.match(/^(|git\+)(https?:.*?)(|.git)$/))) {
-    return url[2];
+  if ((url = repository.match(/^(?:git\+)?(https?:.*?)(?:\.git)?$/))) {
+    return url[1];
   }
 
   // git://github.com/auth0/lock(.git)
-  if ((url = repository.match(/^git:\/\/(.*?)(|.git)$/))) {
-    return `https://${url[2]}`;
+  if ((url = repository.match(/^git:\/\/(.*?)(?:.git)?$/))) {
+    return `https://${url[1]}`;
   }
 
   // github:user/repo
--- a/packages/worker/src/npmRegistry/api.ts
+++ b/packages/worker/src/npmRegistry/api.ts
@@ -7,6 +7,16 @@ const REGISTRY_URL = 'https://replicate.npmjs.com';
 const NPM_API_ORIGIN = 'https://api.npmjs.org';
 
 type DocumentScope = {
+  description?: string;
+  license?: string;
+  homepage?: string;
+  repository?:
+    | string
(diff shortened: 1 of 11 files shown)


packages/worker/src/npmRegistry/api.ts
```diff
@@ -60,3 +90,51 @@ export async function fetchPackageStatsAndMetadata(name: string) {
...
+  }
+
+  // (git+)https://github.com/facebook/react(.git)
+  if ((url = repository.match(/^(|git\+)(https?:.*?)(|.git)$/))) {
+    return url[2];
+  }
+
+  // git://github.com/auth0/lock(.git)
+  if ((url = repository.match(/^git:\/\/(.*?)(|.git)$/))) {
+    return `https://${url[2]}`;
```
ctizen, line 120: Did you mean `url[1]` here?

packages/worker/src/npmRegistry/api.ts
```diff
@@ -60,3 +90,51 @@ export async function fetchPackageStatsAndMetadata(name: string) {
...
+
+function getRepositoryUrl(repository?: string) {
+  let url;
+
+  if (!repository) {
+    return;
+  }
+
+  // (git+)https://github.com/facebook/react(.git)
+  if ((url = repository.match(/^(|git\+)(https?:.*?)(|.git)$/))) {
```
ctizen, line 114: Some considerations for the regexps here and below:
- You could use non-capturing groups (`(?:...)`) which don't get into the output array. 
- You didn't escape the dot metasymbol in the last group.
- Constructs like `(|smth)` are confusing for me, I'd prefer it to be `(smth)?`, and considering first advice about non-capturing groups, `(?:smth)?`.
- The proper equivalent of the expression with fixes above included would be something like `/^(?:git\+)?(https?:.*?)(?:\.git)?$/`

ctizen: Disclaimer: all written above was not tested at all :) https://regex101.com for the rescue.

packages/worker/src/npmRegistry/api.ts
```diff
@@ -60,3 +90,51 @@ export async function fetchPackageStatsAndMetadata(name: string) {
...
+  try {
+    url = new URL(maybeUrl);
+  } catch (_) {
+    return false;
+  }
+
+  return url.protocol === 'http:' || url.protocol === 'https:';
+}
+
+function getRepositoryUrl(repository?: string) {
```
ctizen, line 106: Probably there should be some tests covering this function.

ctizen (changes requested):

ctizen (approved):

packages/worker/src/npmRegistry/api.ts
```diff
@@ -60,3 +90,51 @@ export async function fetchPackageStatsAndMetadata(name: string) {
...
+
+export function getRepositoryUrl(repository?: string) {
+  let url;
+
+  if (!repository) {
+    return;
+  }
+
+  // (git+)https://github.com/facebook/react(.git)
+  if ((url = repository.match(/^(?:git\+)?(https?:.*?)(?:\.git)?$/))) {
```
zardak, line 114: nit: When working with capture groups, destructuring can bring clarity to what's being extracted, like:
```
const urlMatch = repository.match(/^(?:git\+)?(https?:.*?)(?:\.git)?$/);
if (urlMatch) {
  const [_, reporsitoryUrl] = urlMatch;
  ...
}
```

zardak (approved):

packages/worker/src/npmRegistry/api.ts
spalt08: good point, fixed

PR #27, opened 2022-05-29
feat: new UI features for the package list

spalt08:


packages/web/src/components/ui/Package/Package.tsx
```diff
@@ -19,19 +27,82 @@ export type Props = {
...
+  let byteUnitIndex = 0;
+  let sizeInUnits = size;
+
+  while (sizeInUnits > 1024) {
+    sizeInUnits /= 1024;
+    byteUnitIndex += 1;
+  }
+
+  let readableSize = sizeInUnits.toFixed(1);
+  if (readableSize.slice(-2) === '.0') {
```
ctizen, line 95: `.endsWith()`?

packages/web/src/components/ui/Package/Package.tsx
```diff
@@ -19,19 +27,82 @@ export type Props = {
...
+          {repositoryUrl && (
+            <a
+              href={repositoryUrl}
+              target='_blank'
+              rel='noopener noreferrer'
+              className={styles.externalLink}
+            >
+              <Github />
+            </a>
+          )}{' '}
```
ctizen, line 49: This space will appear always, which is probably not the desired behavior. This should be done with css, I guess.

packages/web/src/components/ui/Package/Package.tsx
zardak: There's also a more common but less readable approach of `parseFloat(float.toFixed(x))` which does the trick

zardak (approved):

packages/web/src/components/ui/Package/Package.tsx
spalt08: Fixed

packages/web/src/components/ui/Package/Package.tsx
spalt08: Fixed

ctizen (approved):

PR #30, opened 2022-06-02
feat: package vulnerabilities sourcing

zardak:
This PR adds sourcing of verified npm package vulnerabilities from github advisory db


packages/worker/src/tasks/syncPackageVulnerabilities.ts
```diff
@@ -0,0 +1,140 @@
...
+  PackageVulnerability,
+  'packageName' | 'packageVersionRange' | 'osvId' | 'osvData'
+>;
+
+async function drainStreamToBuffer(stream: Stream.Readable): Promise<Buffer> {
+  return new Promise((resolve, reject) => {
+    const dataChunks: any[] = [];
+
+    stream.on('data', (chunk) => dataChunks.push(chunk));
+    stream.on('end', () => resolve(Buffer.concat(dataChunks)));
```
ctizen, line 22: Won't it leak memory here and below? Actually, I'm not sure about removing listeners from the stream here, but I thought it's worth mentioning it :)

packages/worker/src/tasks/syncPackageVulnerabilities.ts
```diff
@@ -0,0 +1,140 @@
...
+      // Stream has to drained to continue
+      stream.on('end', () => next());
+      stream.resume();
+      return;
+    }
+
+    const fileBuffer = await drainStreamToBuffer(stream);
+    const fileContents = fileBuffer.toString();
+
+    const osvData = JSON.parse(fileContents) as OpenSourceVulnerability;
```
ctizen, line 126: Proposal: use some validation tool (e.g. https://github.com/gcanti/io-ts) on JSON responses from external resources across the project. It would give us some meaningful errors instead of 'access to property of undefined' in case of external API format changes. Not to be done in this PR, but in a separate task maybe.

spalt08:
I was wondering what happens if the package source code is not hosted on github (for example, on gitlab or bitbucket instead)? Will we be able to pull vulnerabilities for such packages?

packages/shared/src/database/migrations/1648763750720-Initial.ts
```diff
@@ -28,7 +28,7 @@ export class Initial1648763750720 implements MigrationInterface {
   }
 
   public async down(queryRunner: QueryRunner): Promise<void> {
-    queryRunner.query(`
+    await queryRunner.query(`
```
spalt08, line 31: 👍

packages/shared/src/utils/env.ts
```diff
@@ -8,6 +8,9 @@ export enum Env {
   AwsRegion = 'AWS_REGION',
   SqsWorkerQueueUrl = 'SQS_WORKER_QUEUE_URL',
 
+  // GitHub
+  GitHubAccessToken = 'GITHUB_ACCESS_TOKEN',
```
spalt08, line 12: It is probably worth adding this variable to the `checkRequiredEnvironmentVariables` payload in the `worker/src/index.ts` file

packages/worker/src/tasks/syncPackageVulnerabilities.test.ts
```diff
@@ -0,0 +1,57 @@
...
+        packageName: 'eventsource',
+        packageVersionRange: '>=0 <2.0.2',
+      },
+    ]);
+
+    await syncPackageVulnerabilities();
+    expect(fetchAdvisoryDbSnapshotMock).toBeCalledTimes(2);
+
+    const updatedSyncedVulnerabilities = await getRepository(PackageVulnerability).find();
+    expect(updatedSyncedVulnerabilities).toMatchObject([
```
spalt08, line 44: JFYI, since we are using a single DB both for the `dev` environment and testing, it might be useful to write tests by keeping a non-empty database in mind. In this case it can be `toContain` instead of `toMatchObject`

packages/worker/src/tasks/syncPackageVulnerabilities.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    if (entityBuffer.length >= batchSize) {
+      await flush();
+    }
+  };
+}
+
+export async function syncPackageVulnerabilities() {
+  const packageVulnerabilitySynchronizer = createPackageVulnerabilitySynchronizer();
+  const extractStream = TarStream.extract();
```
spalt08, line 108: Why don't you use an `unpack` function from the `shared` package instead?

spalt08: Something like
```
const advisoriesPacked = await fetch(githubReqPayload);
const entries = unpack(advisoriesPacked)
const files = Object.keys(entries)
```

packages/worker/src/tasks/syncPackageVulnerabilities.ts
```diff
@@ -0,0 +1,140 @@
...
+    await packageVulnerabilitySynchronizer(osvData);
+
+    next();
+  });
+
+  const dbSnapshotStream = await fetchAdvisoryDbSnapshot();
+
+  const pipelineResult = await pipeline(dbSnapshotStream, gunzip(), extractStream);
+
+  await packageVulnerabilitySynchronizer(null);
```
spalt08, line 137: This line looks quite ugly, as for me

packages/worker/src/tasks/syncPackageVulnerabilities.ts
```diff
@@ -0,0 +1,140 @@
...
+    stream.resume();
+  });
+}
+
+function createPackageVulnerabilitySynchronizer(batchSize = 10) {
+  let entityBuffer: PackageVulnerabilityData[] = [];
+
+  const flush = async () => {
+    await PackageVulnerability.upsert(entityBuffer, {
+      skipUpdateIfNoValuesChanged: true,
```
spalt08, line 34: 👍

spalt08 (changes requested):

packages/worker/src/tasks/syncPackageVulnerabilities.ts
```diff
@@ -0,0 +1,140 @@
...
+    );
+
+    if (entityBuffer.length >= batchSize) {
+      await flush();
+    }
+  };
+}
+
+export async function syncPackageVulnerabilities() {
+  const packageVulnerabilitySynchronizer = createPackageVulnerabilitySynchronizer();
```
spalt08, line 107: The `packageVulnerabilitySynchronizer` name looks like a name for an object. Can we rename it to something like `sync...`?

zardak:
> I was wondering what happens if the package source code is not hosted on github (for example, on gitlab or bitbucket instead)? Will we be able to pull vulnerabilities for such packages?

Yes, github advisory db doesn't exclude packages not hosted on github itself.

packages/shared/src/utils/env.ts
spalt08: Up to you, but I prefer to be sure that the worker application works with any task if it is successfully deployed

ctizen (approved):

PR #34, opened 2022-06-09
Style: sync linters & typecheckers

ctizen:


packages/public-api/tsconfig.build.json
```diff
@@ -2,7 +2,10 @@
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "outDir": "./build",
-    "module": "commonjs"
+    "module": "commonjs",
+    "paths": {
+      "@gradejs-public/shared": ["../shared/src"]
```
spalt08, line 7: It'll probably break the deploy logic. We should use path aliases only in the development mode (specified in `tsconfig.json` in the root). Production configs should not use path aliases for monorepo packages

packages/web/src/index.tsx
```diff
@@ -3,6 +3,7 @@ import ReactGTM from 'react-gtm-module';
 import ReactDOM from 'react-dom';
 import { BrowserRouter } from 'react-router-dom';
 import App from 'components/App';
+// eslint-disable-next-line import/extensions
```
spalt08, line 6: This rule may be quite inconvenient for the front-end package

packages/worker/tsconfig.build.json
```diff
@@ -2,7 +2,10 @@
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "outDir": "./build",
-    "baseUrl": "./src"
+    "baseUrl": "./src",
+    "paths": {
+      "@gradejs-public/shared": ["../../shared/src"]
```
spalt08, line 7: The same here

spalt08 (changes requested):

packages/web/src/index.tsx
ctizen: Good point, fixed in much more proper way.

packages/worker/tsconfig.build.json
ctizen: Good point, fixed in much more proper way.

spalt08 (approved):

PR #36, opened 2022-06-10
feat: analytics tracker for plausible & GA

ctizen:


spalt08 (approved):
LGTM

packages/web/src/services/analytics.ts
```diff
@@ -0,0 +1,87 @@
...
+import ReactGA from 'react-ga';
+
+export type CustomEventProperties = {
+  label?: string;
+  value?: number;
+};
+
+let plausible: ReturnType<typeof Plausible>;
+
+function trackHistory() {
```
zardak, line 11: Isn't there a way to plug into react-router? This seems to be quite a hackish approach.

zardak (approved):

PR #40, opened 2022-08-10
feat: add local startup helpers

ctizen:
Some helper scripts to run local setup in one command

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,9 +21,9 @@
     "build:worker": "yarn workspace @gradejs-public/worker run build",
     "build:public-api": "yarn workspace @gradejs-public/public-api run build",
     "build:backend": "yarn build:shared && yarn build:public-api && yarn build:worker",
-    "dev": "bash cli/local_start.sh",
-    "worker:local:syncPackages": "curl -d '{\"type\":\"syncPackageIndex\"}' -H \"Content-Type: application/json\" -X POST http://localhost:8084",
-    "worker:local:syncVulnerabilities": "curl -d '{\"type\":\"syncPackageVulnerabilities\"}' -H \"Content-Type: application/json\" -X POST http://localhost:8084"
+    "dev:start": "bash cli/local_start.sh",
+    "dev:worker:syncPackages": "curl -d '{\"type\":\"syncPackageIndex\"}' -H \"Content-Type: application/json\" -X POST http://localhost:8084",
+    "dev:worker:syncVulnerabilities": "curl -d '{\"type\":\"syncPackageVulnerabilities\"}' -H \"Content-Type: application/json\" -X POST http://localhost:8084"
   },
   "devDependencies": {
     "@types/jsonpath": "^0.2.0",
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -20,7 +20,10 @@
     "build:shared": "yarn workspace @gradejs-public/shared run build",
     "build:worker": "yarn workspace @gradejs-public/worker run build",
     "build:public-api": "yarn workspace @gradejs-public/public-api run build",
-    "build:backend": "yarn build:shared && yarn build:public-api && yarn build:worker"
+    "build:backend": "yarn build:shared && yarn build:public-api && yarn build:worker",
+    "dev": "bash cli/local_start.sh",
+    "worker:local:syncPackages": "curl -d '{\"type\":\"syncPackageIndex\"}' -H \"Content-Type: application/json\" -X POST http://localhost:8084",
```
spalt08, line 25: I would rather have this in the readme, instead of creating a separate command for each task

packages/web/webpack.config.js
```diff
@@ -8,6 +9,16 @@ const CopyPlugin = require('copy-webpack-plugin');
...
 const distDir = 'dist';
 
+const envDefaults = {
+  API_ORIGIN: 'https://staging.api.gradejs.com',
+  GA_ID: '',
+  PLAUSIBLE_DOMAIN: '',
+  DUMP_ANALYTICS: '',
+};
+const envPath = path.resolve(__dirname, '.env');
+const env = existsSync(envPath) ? dotenv.parse(readFileSync(envPath, { encoding: 'utf-8' })) : {};
```
spalt08, line 19: I would like to avoid parsing env files and having the default value for `API_ORIGIN`, since:
- Production building doesn't have any `.env` files. 
- dotenv can be used specifically for the development environment with a separate command in the `package.json` or inside `local_start.sh`
- Any deploy should trigger an error if the `API_ORIGIN` variable is not provided, just to avoid side-effects

spalt08 (changes requested):

package.json
ctizen: It's just a shortcut. I'd prefer to have a working shortcut for each task than a readme entry (though, _both_ shortcut and readme would be ideal :) ) 

I'll rename it to clarify it's a dev task. If someday we have too many dev tasks in package.json and it will get us nervous, we always can move them to separate bash script or something.

spalt08: Good point, however the logic might break once we start adding worker tasks with custom payloads. 
By having a command per worker task means that we also should update the `package.json` file with all affected changes, which seems redundant for the monorepo

packages/web/webpack.config.js
```diff
@@ -75,18 +79,13 @@ module.exports = (_, argv) => {
...
-        ? []
-        : [
-            new webpack.EnvironmentPlugin({
-              API_ORIGIN: 'https://staging.api.gradejs.com',
-              GA_ID: '',
-              PLAUSIBLE_DOMAIN: '',
-              DUMP_ANALYTICS: '',
-            }),
-          ]),
+      ...(isDevelopment ? [] : [new webpack.EnvironmentPlugin(env)]),
```
spalt08, line 88: It seems like we can use `EnvironmentPlugin(env)` for both development and production

spalt08 (approved):

package.json
ctizen: I don't think this is redundant, as it's a tool for the developer's convenience. Let's leave it here for now.

PR #41, opened 2022-08-10
Feature/rebuild frontend setup

ctizen:
- Add typed api
- Rework pages using redux toolkit


packages/public-api/src/router.ts
```diff
@@ -0,0 +1,73 @@
...
+import { getAffectingVulnerabilities } from './vulnerabilities/vulnerabilities';
+
+const hostnameRe =
+  /^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$|^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)+([A-Za-z]|[A-Za-z][A-Za-z0-9\-]*[A-Za-z0-9])$/;
+
+// created for each request
+export const createContext = (_: CreateExpressContextOptions) => ({}); // no context
+type Context = trpc.inferAsyncReturnType<typeof createContext>;
+
+export const appRouter = trpc
```
spalt08, line 21: Could we split this logic under domain folders?

packages/public-api/src/app.ts
```diff
@@ -1,20 +1,22 @@
...
-  app.use(parseJson);
-  app.use(healthCheckRouter);
-  app.use(websiteRouter);
+  if (process.env.NODE_ENV !== 'test') {
+    app.use(cors);
+  }
+
+  app.use(
+    '/',
+    trpcExpress.createExpressMiddleware({
```
spalt08, line 15: What's the purpose of the wildcard import here? Can't we just use `createExpressMiddleware` imported directly?

packages/public-api/src/app.ts
```diff
@@ -1,20 +1,22 @@
...
+import { appRouter, createContext } from './router';
 
 export function createApp() {
   const app = express();
 
-  app.use(cors);
-  app.use(parseJson);
-  app.use(healthCheckRouter);
-  app.use(websiteRouter);
+  if (process.env.NODE_ENV !== 'test') {
```
spalt08, line 9: Can we use `!isTest()` shortcut here?

packages/public-api/src/middleware/common.ts
```diff
@@ -1,11 +1,34 @@
...
+import { ZodError } from 'zod';
 import createCorsMiddleware from 'cors';
-import express, { Request, Response, NextFunction } from 'express';
+import { Request, Response, NextFunction } from 'express';
 import { NotFoundError, respondWithError } from './response';
 
-// TODO: add whitelist origins
-export const cors = createCorsMiddleware({ maxAge: 1800 });
-export const parseJson = express.json();
+const whitelist = {
```
spalt08, line 6: Let's rename it as `originWhiteList`.
Please also add `localhost` origin to the staging white list

packages/public-api/src/middleware/common.ts
```diff
@@ -1,11 +1,34 @@
...
+  production: ['https://gradejs.com'],
+  staging: ['https://staging.gradejs.com'],
+  development: ['http://localhost:3000'],
+};
+
+const getNodeEnv = () => {
```
spalt08, line 12: I believe we already have such function in the `shared` package

packages/shared/src/internalApi/api.ts
```diff
@@ -79,6 +79,10 @@ export async function fetchEndpoint<T>(
     }
   }
 
+  if (process.env.NODE_ENV === 'development') {
```
spalt08, line 82: What if I want to run the dev environment locally and use internal API?

packages/web/src/components/layouts/Website/Website.tsx
```diff
@@ -19,39 +18,38 @@ export type Props = {
...
-  packages: DetectedPackageData[];
+  isLoading: boolean;
+  isPending: boolean;
+  packages: WebPagePackage[];
   vulnerabilities: Record<string, PackageVulnerabilityData[]>;
-  webpages: Array<{
-    status: string;
-  }>;
-  onFiltersApply: SubmitHandler<FormData>;
+  webpages: WebPage[];
```
spalt08, line 25: I believe we should strictly use `trpc` types instead of database entities on the client, since some fields data types might be changed (`Date` -> `string` for example)

packages/web/src/services/apiClient.ts
```diff
@@ -0,0 +1,8 @@
+import { createTRPCClient } from '@trpc/client';
+import type { AppRouter } from '../../../public-api/src/router';
+
+export const client = createTRPCClient<AppRouter>({
+  url: process.env.API_ORIGIN!,
```
spalt08, line 5: I would like to avoid `!` here

packages/web/src/store/slices/websiteResults.ts
```diff
@@ -0,0 +1,75 @@
...
+import { trackCustomEvent } from '../../services/analytics';
+
+const defaultDetectionResult = {
+  packages: [] as WebPagePackage[],
+  vulnerabilities: {} as Record<string, PackageVulnerabilityData[]>,
+  webpages: [] as WebPage[],
+};
+
+const initialState = {
+  filters: { ...DefaultFiltersAndSorters },
```
spalt08, line 14: Why can't we have filters state on the component level? I believe we should reset them on every website

packages/web/src/store/slices/websiteResults.ts
```diff
@@ -0,0 +1,75 @@
...
+  });
+
+const hasPendingPages = (result: DetectionResult) =>
+  !!result.webpages.find((item) => item.status === 'pending');
+
+const getWebsite = createAsyncThunk('websiteResults/getWebsite', async (hostname: string) => {
+  const loadStartTime = Date.now();
+  await client.mutation('syncWebsite', hostname);
+  let results = await client.query('getWebsite', hostname);
+  while (hasPendingPages(results)) {
```
spalt08, line 32: What if a user loads a pending page and then goes on another page with results without waiting for the first page?

packages/web/src/components/pages/Home.tsx
```diff
@@ -0,0 +1,46 @@
...
+} from '../../store';
+import { useNavigate } from 'react-router-dom';
+
+export function HomePage() {
+  const navigate = useNavigate();
+  const dispatch = useAppDispatch();
+  const state = useAppSelector(homeDefaultSelector);
+
+  const handleDetectStart = useCallback(async (data: { address: string }) => {
+    trackCustomEvent('HomePage', 'WebsiteSubmitted');
```
spalt08, line 19: Why do we track `WebsiteLoaded` inside redux, but `WebsiteSubmitted` inside a component?

packages/web/src/components/pages/WebsiteResults.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { FiltersState } from '../layouts/Filters/Filters';
+
+export function WebsiteResultsPage() {
+  const { hostname } = useParams();
+  const navigate = useNavigate();
+  const dispatch = useAppDispatch();
+  const { webpages, vulnerabilities } = useAppSelector(selectors.default);
+  const packagesFiltered = useAppSelector(selectors.packagesSortedAndFiltered);
+  const isProtected = useAppSelector(selectors.isProtected);
+  const isPending = useAppSelector(selectors.isPending);
```
spalt08, line 21: Do we really need a separate selector per each flag?

packages/web/src/components/pages/WebsiteResults.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  // TODO: properly handle history/routing
+  useEffect(() => {
+    if (!hostname || isError) {
+      navigate('/', { replace: true });
+    }
+  });
+
+  if (isProtected) {
+    trackCustomEvent('HostnamePage', 'SiteProtected');
```
spalt08, line 41: Tracking events on each re-render seems like a bad practice

spalt08 (changes requested):

packages/shared/src/internalApi/api.ts
```diff
@@ -79,6 +79,10 @@ export async function fetchEndpoint<T>(
...
+    console.log('Request to internal API: ', requestUrl.toString(), requestInit);
```
zardak, line 83: I'd much rather use a logger with loglevels or leave this out entirely, attaching a debugger should solve most of the cases where this'll be useful imo.

packages/web/src/store/slices/websiteResults.ts
zardak: This is a controversial point, there use cases that would benefit from non-resetting filters, we might even want to persist filter state in url query. Let's review this later, during redesign.

packages/web/src/store/slices/websiteResults.ts
```diff
@@ -0,0 +1,75 @@
...
+    await sleep(5000);
+    await client.mutation('syncWebsite', hostname);
+    results = await client.query('getWebsite', hostname);
```
zardak, line 35: Could you clarify why would we need two separate requests in here?

packages/public-api/src/middleware/common.ts
ctizen: The one in `shared` appears to be insufficient for this use. Do you mind replacing that function with this implementation?

packages/shared/src/internalApi/api.ts
ctizen: I didn't get it. How is the question related to the code?

packages/public-api/src/app.ts
ctizen: The reason was that I didn't find a way to support cors requests in supertest when I changed router tests. Googled a bit now and replaced this condition with `.set('Origin', 'http://localhost:3000')` in the tests.

packages/web/src/store/slices/websiteResults.ts
ctizen: Filtering is not a UI thing, it's data related thing. It makes much more sense to keep it in the app state.

packages/web/src/components/pages/Home.tsx
ctizen: That's why I left a TODO there :) Usually event tracking is located in separate redux middleware, but I think we better discuss this.

packages/web/src/store/slices/websiteResults.ts
ctizen: Valid point! Added cleanup code in the component, though we should test it once again after redesign, because currently we don't have any way to get out of results page without page reload.

packages/web/src/components/pages/WebsiteResults.tsx
ctizen: There will be no further rerenders in this condition branch. Also, see my previous comment about tracking events in redux - probably we'll move it from here.

packages/public-api/src/middleware/common.ts
```diff
@@ -1,11 +1,34 @@
...
 import createCorsMiddleware from 'cors';
-import express, { Request, Response, NextFunction } from 'express';
+import { Request, Response, NextFunction } from 'express';
 import { NotFoundError, respondWithError } from './response';
 
-// TODO: add whitelist origins
-export const cors = createCorsMiddleware({ maxAge: 1800 });
-export const parseJson = express.json();
+const whitelist = {
+  production: ['https://gradejs.com'],
```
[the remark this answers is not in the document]
ctizen: 🤔 sounds legit actually. I need to check out if comma-separated hosts are a good choice for this parameter

packages/shared/src/internalApi/api.ts
ctizen: I tend to agree with this. Ok, let's remove this line

packages/web/src/store/slices/websiteResults.ts
ctizen: Changed this, but I think we should take a look at it again, it still looks weird.

packages/public-api/src/middleware/common.ts
spalt08: Yeah, that should work

packages/shared/src/internalApi/api.ts
spalt08: nvm, I guess you can run development environment with `NODE_ENV=staging`

packages/web/src/components/pages/WebsiteResults.tsx
spalt08: Please leave a TODO

packages/web/src/store/slices/websiteResults.ts
zardak: Looks passable to me for now, we'll revisit it with redesign impl

packages/public-api/src/middleware/common.ts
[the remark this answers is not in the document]
ctizen: done

spalt08 (approved):

zardak (approved):

PR #43, opened 2022-08-19
feat: Implement server-side rendering

ctizen:
Fixes #181


cli/archive_artifacts.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# Strip dev packages
+yarn install --production --pure-lockfile
+
+tar cf build.tar \
```
spalt08, line 6: Can you please add a comment somewhere mentioning that we need `tar` in order to keep monorepo symlinks

packages/web/src/server.tsx
```diff
@@ -0,0 +1,33 @@
+import express from 'express';
+import React from 'react';
+import { Provider } from 'react-redux';
+import ReactDOMServer from 'react-dom/server';
+import { StaticRouter } from 'react-router-dom/server';
+import { getPort, getClientVars } from '@gradejs-public/shared/src/utils/env';
```
ctizen, line 6: Not sure how to fix this (and if we should, actually). Webpack can't figure this out with tree-shaking, because the trouble comes from the build process, not from the bundle startup:
```
ERROR in ../../node_modules/pg/lib/native/client.js 4:13-33
Module not found: Error: Can't resolve 'pg-native' in '/Users/olegklimenko/projects/gradejs/node_modules/pg/lib/native'
resolve 'pg-native' in '/Users/olegklimenko/projects/gradejs/node_modules/pg/lib/native'
  Parsed request is a module
  using description file: /Users/olegklimenko/projects/gradejs/node_modules/pg/package.json (relative path: ./lib/native)
    resolve as module
      looking for modules in /Users/olegklimenko/projects/gradejs/packages/web/src
        single file module
          using description file: /Users/olegklimenko/projects/gradejs/packages/web/package.json (relative path: ./src/pg-native)
            no extension
              /Users/olegklimenko/projects/gradejs/packages/web/src/pg-native doesn't exist
            .ts
              /Users/olegklimenko/projects/gradejs/packages/web/src/pg-native.ts doesn't exist
            .tsx
              /Users/olegklimenko/projects/gradejs/packages/web/src/pg-native.tsx doesn't exist
            .js
              /Users/olegklimenko/projects/gradejs/packages/web/src/pg-native.js doesn't exist
        /Users/olegklimenko/projects/gradejs/packages/web/src/pg-native doesn't exist
      /Users/olegklimenko/projects/gradejs/node_modules/pg/lib/native/node_modules doesn't exist or is not a directory
      /Users/olegklimenko/projects/gradejs/node_modules/pg/lib/node_modules doesn't exist or is not a directory
      /Users/olegklimenko/projects/gradejs/node_modules/pg/node_modules doesn't exist or is not a directory
      /Users/olegklimenko/projects/gradejs/node_modules/node_modules doesn't exist or is not a directory
      looking for modules in /Users/olegklimenko/projects/gradejs/node_modules
        single file module
          using description file: /Users/olegklimenko/projects/gradejs/package.json (relative path: ./node_modules/pg-native)
            no extension
              /Users/olegklimenko/projects/gradejs/node_modules/pg-native doesn't exist
            .ts
              /Users/olegklimenko/projects/gradejs/node_modules/pg-native.ts doesn't exist
            .tsx
              /Users/olegklimenko/projects/gradejs/node_modules/pg-native.tsx doesn't exist
            .js
              /Users/olegklimenko/projects/gradejs/node_modules/pg-native.js doesn't exist
        /Users/olegklimenko/projects/gradejs/node_modules/pg-native doesn't exist
      /Users/olegklimenko/projects/node_modules doesn't exist or is not a directory
      /Users/olegklimenko/node_modules doesn't exist or is not a directory
      /Users/node_modules doesn't exist or is not a directory
      /node_modules doesn't exist or is not a directory
 @ ../../node_modules/pg/lib/native/index.js 2:0-36
 @ ../../node_modules/pg/lib/index.js 29:26-45 40:24-43
 @ ../../node_modules/typeorm/platform/PlatformTools.js 68:27-40
 @ ../../node_modules/typeorm/globals.js 6:22-57
 @ ../../node_modules/typeorm/index.js 11:26-46
 @ ../shared/build/database/connection.js 8:18-36
 @ ../shared/build/index.js 30:13-45
 @ ./src/server.tsx 8:17-50

webpack 5.70.0 compiled with 1 error and 24 warnings in 7240 ms
✨  Done in 9.64s.
```

The error can be worked around using something like this diff:
```diff
diff --git a/packages/web/src/server.tsx b/packages/web/src/server.tsx
index 25c992c..389e279 100644
--- a/packages/web/src/server.tsx
+++ b/packages/web/src/server.tsx
@@ -3,7 +3,7 @@ import React from 'react';
 import { Provider } from 'react-redux';
 import ReactDOMServer from 'react-dom/server';
 import { StaticRouter } from 'react-router-dom/server';
-import { getPort, getClientVars } from '@gradejs-public/shared/src/utils/env';
+import { getPort, getClientVars } from '@gradejs-public/shared';
 import { store } from './store';
 import { App } from './components/App';
 import path from 'path';
diff --git a/packages/web/webpack/client.ts b/packages/web/webpack/client.ts
index 0361c08..7fba533 100644
--- a/packages/web/webpack/client.ts
+++ b/packages/web/webpack/client.ts
@@ -12,7 +12,7 @@ export const clientConfig: (mode: 'development' | 'production', watch: boolean)
   watch
 ) => ({
   entry: join(__dirname, '..', srcDir, 'index.tsx'),
-  ...configCommon(mode),
+  ...configCommon(mode, false),
   module: {
     rules: [
       {
diff --git a/packages/web/webpack/common.ts b/packages/web/webpack/common.ts
index b8f495f..0417e35 100644
--- a/packages/web/webpack/common.ts
+++ b/packages/web/webpack/common.ts
@@ -6,11 +6,15 @@ const SpritePlugin = require('svg-sprite-loader/plugin');

 export const srcDir = 'src';

-export const configCommon: (mode: 'development' | 'production') => Configuration = (mode) => ({
+export const configCommon: (
+  mode: 'development' | 'production',
+  isServer: boolean
+) => Configuration = (mode, isServer) => ({
   mode,
   resolve: {
     extensions: ['.ts', '.tsx', '.js'],
     modules: [resolve(__dirname, '..', srcDir), 'node_modules'],
+    ...(isServer ? { alias: { 'pg-native': false } } : {}),
   },
   devtool: mode === 'development' ? 'eval-source-map' : false,
 });
diff --git a/packages/web/webpack/server.ts b/packages/web/webpack/server.ts
index c31830e..5ae9751 100644
--- a/packages/web/webpack/server.ts
+++ b/packages/web/webpack/server.ts
@@ -10,7 +10,7 @@ export const serverConfig: (mode: 'development' | 'production', watch: boolean)
   watch
 ) => ({
   entry: join(__dirname, '..', srcDir, 'server.tsx'),
-  ...configCommon(mode),
+  ...configCommon(mode, true),
   module: {
     rules: [
       {

```

But still, webpack tree-shaking does not work properly in these conditions: the resulting bundle is around 3x-4x of its original size, both with minification and without it. I suppose that the key reason is node modules are mostly in CJS format, while tree-shaking is ESM-oriented.

I'd suggest avoiding `export *`s where possible.

ctizen: Some ideas to check after:
- `webpack-node-externals` package and externalization of imports. Checked now and got a reduced size of the main bundle by ~15%. Though, if I try to import from `@gradejs-public/shared`, the size is still 3x-4x of the original.
- Typescript misconfiguration (e.g. `type: "commonjs"` that we have now). Tried the "es6" setting, but did not get any improvement. Need to check if webpack scripts are processed with tsconfig different from one being used during processing of main bundle.

package.json
```diff
@@ -22,7 +22,7 @@
     "build:public-api": "yarn workspace @gradejs-public/public-api run build",
     "build:backend": "yarn build:shared && yarn build:public-api && yarn build:worker",
     "dev:start": "bash cli/local_start.sh",
-    "dev:start:onlyWeb": "API_ORIGIN=https://api.staging.gradejs.com npm start --prefix packages/web"
+    "dev:start:onlyWeb": "API_ORIGIN=https://api.staging.gradejs.com npm dev:start --prefix packages/web"
```
spalt08, line 25: Why not `dev:start:web`?

packages/web/src/components/App.tsx
```diff
@@ -14,7 +12,11 @@ const locationChangeHandler = initAnalytics();
  *   <App />
  * </BrowserRouter>
  */
-export function App() {
+export function App({
```
spalt08, line 15:
```ts
type Props = {
  locationChangeHandler: (url?: string | URL) => void;
}

export function App({ locationChangeHandler }: Props) {
```

packages/web/src/components/layouts/Filters/Filters.tsx
```diff
@@ -30,7 +30,9 @@ export default function Filters({ onSubmit }: Props) {
   const watchFilterByName = watch('filter');
   let hideHandle: () => void;
   useEffect(() => {
-    document.body.addEventListener('click', hideHandle);
+    if (typeof document !== 'undefined') {
```
spalt08, line 33: This part looks quite ugly. Have you considered `useClickOutside` hook?

packages/web/src/server.tsx
spalt08: Can't we just use separate `env` toolkit for the `web` package for now? Also you can import it directly with `../../` and without package name

packages/web/src/services/apiClient.ts
```diff
@@ -18,6 +21,10 @@ if (!process.env.API_ORIGIN) {
 
 export const client = createTRPCClient<AppRouter>({
   url: process.env.API_ORIGIN,
+  fetch: typeof window === 'undefined' ? (f as any) : window.fetch.bind(window),
+  headers: {
+    Origin: process.env.CORS_ORIGIN,
```
spalt08, line 26: I'm not sure if the api need to throw a 500 error and check for origin. A browser has a built-in mechanisms to check CORS.

So I guess the public api implementation should set `Access-Control-Allow-*` headers and do not throw a 500 error

packages/web/webpack/build.ts
```diff
@@ -0,0 +1,31 @@
+import webpack from 'webpack';
+import yargs from 'yargs/yargs';
+import { hideBin } from 'yargs/helpers';
+import { clientConfig } from './client';
+import { serverConfig } from './server';
+
+const args = yargs(hideBin(process.argv))
```
spalt08, line 7: maybe `cli/build.ts` ?

packages/web/webpack/client.ts
```diff
@@ -0,0 +1,90 @@
...
+  },
+  plugins: [
+    ...pluginsCommon(mode, false),
+    new HtmlWebpackPlugin({
+      template: join(__dirname, '..', 'index.html'),
+    }),
+    new CopyPlugin({
+      patterns: [
+        { from: 'robots.txt', to: 'robots.txt' },
+        { from: 'src/assets/sharing-image.png', to: 'sharing-image.png' },
```
spalt08, line 79: `to: 'assets/sharing-image'`

packages/web/webpack/server.ts
```diff
@@ -0,0 +1,66 @@
...
+        exclude: /\.module.scss$/,
+        use: [MiniCssExtractPlugin.loader, 'css-loader', 'sass-loader'],
+      },
+    ],
+  },
+  plugins: pluginsCommon(mode, true),
+  output: {
+    filename: '[name].js',
+    path: join(__dirname, '..', distDir),
+    libraryTarget: 'umd',
```
spalt08, line 58: why umd?

packages/web/src/services/apiClient.ts
ctizen: Probably this is not related, too, though fine-tuning of error handling on public-api end might be good idea.

packages/web/webpack/build.ts
ctizen: This is the only cli script in the web package, so I wouldn't make a separate `cli` folder for it, at least for now

packages/web/webpack/client.ts
ctizen: We don't have a separate `assets` folder now inside `dist/static`.

packages/web/src/server.tsx
ctizen: Ok, imported by relative path and left a TODO item

packages/web/webpack/server.ts
ctizen: Not sure what type should I select here, as there is actually no difference. Let's stick to `commonjs2` setting.

packages/web/webpack/client.ts
spalt08: It should be `static/sharing-image.png` than
https://gradejs.com/static/sharing-image.png

spalt08 (approved):

packages/web/webpack/common.ts
```diff
@@ -0,0 +1,28 @@
...
+    extensions: ['.ts', '.tsx', '.js'],
+    modules: [resolve(__dirname, '..', srcDir), 'node_modules'],
+  },
+  devtool: mode === 'development' ? 'eval-source-map' : false,
+});
+
+export const pluginsCommon = (mode: string, isServer: boolean) => [
+  new DefinePlugin({
+    __isServer__: isServer,
+    ...(isServer ? {} : {'process.env': 'window.env'}),
```
zardak, line 21: We can always refer to `process.env` global, while assigning env vars as `window.process = { env: {...} }` client-side. IMO, this is same amount of magic, but less clutter overall.

PR #47, opened 2022-08-25
fix: 500 error on public api healthcheck

ctizen:


packages/public-api/src/app.ts
```diff
@@ -5,6 +5,8 @@ import { appRouter, createContext } from './router';
 
 export function createApp() {
   const app = express();
+  app.get('/', (_, res) => res.send('gradejs-public-api'));
+  app.get('/healthcheck', (_, res) => res.send('gradejs-public-api'));
```
spalt08, line 9: Why do we need a second endpoint?

ctizen: ok, removed

spalt08 (approved):

PR #48, opened 2022-08-26
feat: completely redesigned product

spalt08:


packages/web/src/components/layouts/HomeNew/HomeNew.tsx
```diff
@@ -0,0 +1,160 @@
+import React from 'react';
+import Hero from '../../ui/Hero/Hero';
+import Container from 'components/ui/Container/Container';
+import CardList from '../../ui/CardList/CardList';
+import CardGroups from '../../ui/CardGroups/CardGroups';
+import Footer from '../../ui/Footer/Footer';
+import CardGroup from '../../ui/CardGroup/CardGroup';
+
+export type HomeNewProps = {
```
spalt08, line 9: For a single component inside a single file I would rather use `Props` name always without an export. The naming would be simpler and we should avoid importing types from components.

```ts
type Props = { /* Component props */ }

export default function Component(props: Props) {
  // ...
}
```

packages/web/src/components/layouts/HomeNew/HomeNew.tsx
```diff
@@ -0,0 +1,160 @@
...
+  suggestions: string[];
+};
+
+export default function HomeNew({ suggestions }: HomeNewProps) {
+  const popularCards = [
```
spalt08, line 14: I would rather define constants inside storybook stories files and pass them as component properties. 

Any component should not contain any hard-coded constants.

packages/web/src/components/ui/Card/Card.module.scss
```diff
@@ -0,0 +1,231 @@
...
+  display: flex;
+  align-items: center;
+}
+
+.icon {
+  flex-shrink: 0;
+  margin-right: 16px;
+  display: inline-flex;
+
+  img {
```
spalt08, line 36: I would like to avoid any tag name selectors. Let's use only class names and pseudo elements.

packages/web/src/components/ui/Card/Card.module.scss
```diff
@@ -0,0 +1,231 @@
...
+  }
+}
+
+.tags {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+  justify-content: flex-start;
+
+  & > * {
```
spalt08, line 79: Let's avoid `*` selectors as well. Since CSS selectors are read by a browser from right to left, this selector will be checked for any html element on a page.

packages/web/src/utils/helpers.tsx
```diff
@@ -0,0 +1,3 @@
+export function numberWithSpaces(x: number) {
```
spalt08, line 1: Should probably be:
```
function formatNumber(x) {
  return x.toLocaleString()
}
```

packages/web/src/components/ui/CardList/CardList.tsx
```diff
@@ -0,0 +1,21 @@
...
+import Card, { CardProps } from '../Card/Card';
+import clsx from 'clsx';
+
+export type CardListProps = {
+  cards: CardProps[];
+  variant?: 'default' | 'vertical' | string;
+};
+
+export default function CardList({ cards, variant = 'default' }: CardListProps) {
+  const style = { '--total': cards.length } as React.CSSProperties;
```
spalt08, line 12: I'm not sure about css variables inside `style` attribute. @zardak @ctizen what do you think?

spalt08 (commented):
Feel free to replace existing elements, rather than `ElementNew`, since we don't need to keep legacy elements

spalt08:
@ctizen @zardak it seems like the layout should be scaled down to ~80%. Currently the main page looks quite ugly.

packages/web/src/components/ui/CardList/CardList.tsx
KnoWabLe: In order for this horizontal scrollable grid container on mobile to work it needs an external total items counter coming from JS, so it can properly calculate `grid-template-columns`. [Working example and reference](https://codepen.io/dannievinther/pen/PdvvQv)

[CSS-only example](https://codepen.io/dannievinther/pen/PdvvQv) leaves huge gap at the end of container:
image

packages/web/src/components/ui/Hero/Hero.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { Arrow } from '../../icons';
+import HeaderNew from '../HeaderNew/HeaderNew';
+
+export type HeroProps = {
+  inputText?: string;
+  loading?: boolean;
+  suggestions: string[];
+};
+
+export default function Hero({ inputText, suggestions, loading = false }: HeroProps) {
```
ctizen, line 14: Is this some common name? Sounds misleading to me, never seen this before.

packages/web/src/components/ui/Card/Card.tsx
```diff
@@ -0,0 +1,127 @@
...
+            </div>
+          )}
+
+          <div className={styles.title}>{title}</div>
+        </header>
+
+        {description && <div className={styles.description}>{description}</div>}
+      </div>
+
+      {variant === 'toAll' && (
```
ctizen, line 52: Is it `to-all` or `toAll`? Also could you please remove `string` type from `variant` field to avoid such mistakes.

packages/web/src/components/ui/CardList/CardList.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import styles from './CardList.module.scss';
+import Card, { CardProps } from '../Card/Card';
+import clsx from 'clsx';
+
+export type CardListProps = {
+  cards: CardProps[];
+  variant?: 'default' | 'vertical' | string;
```
ctizen, line 8: Same here. Please remove `| string` everywhere in such cases.

packages/web/src/components/ui/CardList/CardList.tsx
ctizen: Is grid even required here? Looks like it can be implemented with a plain container with fixed-width inline blocks within.

packages/web/src/components/ui/Footer/Footer.tsx
```diff
@@ -0,0 +1,49 @@
...
+              target='_blank'
+              rel='norefferer noreferrer'
+              className={styles.link}
+            >
+              About
+            </a>
+            <a
+              href='https://github.com/gradejs/gradejs/discussions'
+              target='_blank'
+              rel='norefferer noreferrer'
```
ctizen, line 23: What is `noreFFerer`? Was it intended as `noopener noreferrer`?

packages/web/src/components/ui/HeaderNew/HeaderNew.tsx
```diff
@@ -0,0 +1,44 @@
...
+import styles from './HeaderNew.module.scss';
+import Container from '../Container/Container';
+import { Github, Logo } from 'components/icons';
+import { trackCustomEvent } from '../../../services/analytics';
+
+export default function HeaderNew() {
+  return (
+    <Container>
+      <header className={styles.header}>
+        <Logo className={styles.logo} />
```
ctizen, line 11: I think we should not pass external class names into components. If we need to change component representation, we better define some variants and pass constant here

ctizen:
> @ctizen @zardak it seems like the layout should be scaled down to ~80%. Currently the main page looks quite ugly.

Probably we should ask design team about it. E.g. I'd scale header to 80%, but cards to 60%.

packages/web/src/components/ui/Hero/Hero.tsx
KnoWabLe: Quite common, actually.

Few examples:
github
image

audible
image

vercel
image

packages/web/src/components/ui/Footer/Footer.tsx
KnoWabLe: Didn't notice. Copied from current production Header links https://github.com/gradejs/gradejs/blob/develop/packages/web/src/components/ui/Header/Header.tsx#L13

packages/web/src/components/ui/CardList/CardList.tsx
KnoWabLe: Yep. Flexbox solution would be sufficient enough.

packages/web/src/components/ui/Hero/Hero.tsx
ctizen: Ok, thanks :) Another day, another interesting fact

packages/web/src/components/layouts/HomeNew/HomeNew.tsx
zardak: Seems like this is the place where data fetching is supposed to happen, I'd just leave it as-is and revisit once the PR is past draft stage.

packages/web/src/components/layouts/HomeNew/HomeNew.tsx
```diff
@@ -0,0 +1,160 @@
...
+  ];
+
+  return (
+    <>
+      <Hero suggestions={suggestions} />
+
+      <Container>
+        <CardGroups>
+          <CardGroup title='Popular search queries'>
+            <CardList cards={popularCards} />
```
zardak, line 140: nit: We'll probably need separate components per variant (i.e. `<PopularWebsiteCard>` `<PopularPackageCard>`). This seems to align with how styles are currently structured for the cards. just taking note to fix this after drafting phase.

packages/web/src/components/ui/Card/Card.tsx
```diff
@@ -0,0 +1,127 @@
...
+              stroke='#212121'
+              strokeWidth='2'
+              strokeLinecap='round'
+              strokeLinejoin='round'
+            />
+          </svg>
+        </button>
+      )}
+
+      {packageTags && (
```
zardak, line 73: Seems like this and below would be better off with a separate react component passed in as a prop.

packages/web/src/components/ui/CardList/CardList.tsx
zardak: > I'm not sure about css variables inside style attribute.

I'd say this is not a bad practice per-se, if used sparingly. However, we should try solving this specific case with flexbox, unless we need to snap the scroll to items.

packages/web/src/components/ui/CardList/CardList.tsx
```diff
@@ -0,0 +1,21 @@
...
+};
+
+export default function CardList({ cards, variant = 'default' }: CardListProps) {
+  const style = { '--total': cards.length } as React.CSSProperties;
+
+  return (
+    <div className={clsx(styles.grid, styles[variant])} style={style}>
+      {cards.map((card) => (
+        <Card key={card.title} {...card} />
```
zardak, line 17: nit: Items rendered here will have an `id`, and should be keyed by it.

packages/web/src/components/ui/HeaderNew/HeaderNew.tsx
zardak: Simply adding a wrapper would also suffice, if I'm not mistaken.

packages/web/src/components/ui/HeaderNew/HeaderNew.tsx
```diff
@@ -0,0 +1,44 @@
...
+        <div className={styles.nav}>
+          <a
+            href='https://github.com/gradejs/gradejs/discussions/6'
+            target='_blank'
+            rel='norefferer noreferrer'
+            className={styles.navLink}
+            onClick={() => trackCustomEvent('ClickExternalLink', 'About')}
```
zardak, line 18: nit: Such functions should be memoized, once we're past draft phase.

packages/web/src/components/ui/Card/Card.module.scss
zardak: This is potentially true for this case, but I'd advise against banning usage of `*` selectors, we'll likely to end up with a lot of unneeded complexity in react-side of markup.

spalt08: I can't think about any useful `*` use case other than a single usage in a `global.css` for things like `box-sizing`

packages/web/src/components/ui/Footer/Footer.tsx
KnoWabLe: Nowadays probably should be just `<a href="..." rel="noreferrer" target="_blank">`. Ref: [https://html.spec.whatwg.org/multipage/links.html#link-type-noreferrer](https://html.spec.whatwg.org/multipage/links.html#link-type-noreferrer)

packages/web/.storybook/main.js
```diff
@@ -16,7 +16,9 @@ module.exports = {
         {
           loader: 'css-loader',
           options: {
-            modules: true,
+            modules: {
+              exportLocalsConvention: 'camelCase',
```
spalt08, line 20: Can't we use the camel case selectors always and omit this property?

packages/web/src/components/ui/HeaderNew/HeaderNew.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import styles from './HeaderNew.module.scss';
+import Container from '../Container/Container';
+import { Icon } from '../Icon/Icon';
+import { trackCustomEvent } from '../../../services/analytics';
+
+export default function HeaderNew() {
```
ctizen, line 7: Will it be renamed in future? I think we won't leave old components in the repo, so it could be just 'Header'

KnoWabLe: Yes, it will. For now i'm using `New` suffix to keep old ones alongside as references, especially when it's not just static markup, but some functionality involved.

In case of `Header` it could easily be replaced, because it's mostly static links, so i'll do it in future commits/updates.

spalt08: Let's change existing components instead of adding a suffix

packages/web/src/components/layouts/HomeNew/HomeNew.tsx
```diff
@@ -0,0 +1,176 @@
+import React from 'react';
+import { CardProps } from '../../ui/Card/Card';
+import Hero from '../../ui/Hero/Hero';
+import Container from 'components/ui/Container/Container';
+import CardList from '../../ui/CardList/CardList';
+import CardGroups from '../../ui/CardGroups/CardGroups';
+import Footer from '../../ui/Footer/Footer';
+import CardGroup from '../../ui/CardGroup/CardGroup';
```
ctizen, line 8: A few thoughts about component structure:
- I'd suggest having a simple `index.ts` reexporting the component in each component folder so we could just `import { Component } from '../../ui/Component'`.
- I would also recommend against `export default`ing components in favor of named exports.
- I'd ask for better uniformity of import paths, e.g. import either from `components/ui/...` or from `../../ui/...`, but not both.

PS. Not sure if this is really required to be fixed in this PR, we can do it later actually.

packages/web/src/components/layouts/SearchResults/SearchResults.tsx
```diff
@@ -0,0 +1,337 @@
...
+        id: 'trueadm',
+        name: 'trueadm',
+      },
+    ],
+    featuredItems: ['acdlite', 'gaearon', 'sophiebits', 'trueadm'],
+  };
+
+  const [selectedKeywords, setSelectedKeywords] = useState<string[] | []>([]);
+  const [selectedProblems, setSelectedProblems] = useState<string[] | []>([]);
+  const [selectedAuthors, setSelectedAuthors] = useState<string[] | []>([]);
```
ctizen, line 183: Just a note: in general, you don't need to specify an empty array type explicitly, `[]` is perfectly assignable to `string[]`. 

Anyway, probably I could live with this _here_, because filters will be managed by global app state manager afterwards.

packages/web/src/components/ui/Chip/Chip.tsx
```diff
@@ -0,0 +1,46 @@
...
+import clsx from 'clsx';
+import styles from './Chip.module.scss';
+import { IconProps } from '../Icon/Icon';
+
+export type ChipProps = {
+  children: React.ReactNode;
+  className?: string;
+  variant?: 'primary' | 'secondary' | 'outlined' | 'suggest' | string;
+  size?: 'default' | 'medium' | 'large' | string;
+  font?: 'sans-serif' | 'monospace' | string;
```
ctizen, line 11: I'd suggest removing `| string` from props typings

packages/web/src/components/ui/SidebarCategory/SidebarCategory.tsx
```diff
@@ -0,0 +1,187 @@
...
+    featuredItems: string[];
+  };
+  simpleCategory?: string[];
+  selectedKeywords: string[] | [];
+  selectHandler: (name: string) => void;
+  renderComponent: 'chip' | 'checkbox' | 'person';
+  searchable?: boolean;
+};
+
+export default function SidebarCategory({
```
ctizen, line 37: The whole component looks messy, but I don't have any improvement suggestions right now. Left this comment to grab some attention here in future.

packages/web/src/components/ui/SidebarCategorySearch/SidebarCategorySearch.tsx
```diff
@@ -0,0 +1,104 @@
...
+  selectedItems,
+  renderComponent,
+  selectHandler,
+}: Props) {
+  return (
+    <>
+      <div className={styles.searchWrapper}>
+        {/* FIXME: Don't know why looking glass icon needs to be to the left side and */}
+        {/* disappear if input is not empty. Why not show it on right side only like in */}
+        {/* header search bar and toggle clear icon when not empty, probably design mistake */}
```
ctizen, line 40: Gotta discuss this with the UX crew.

packages/web/src/components/ui/SidebarCategorySearch/SidebarCategorySearch.tsx
```diff
@@ -0,0 +1,104 @@
...
+                {children.map(({ name, id }: listItem) => (
+                  <div
+                    key={id}
+                    className={clsx(
+                      styles.groupItem,
+                      selectedItems.includes(name) && styles.groupItemActive
+                    )}
+                    onClick={() => selectHandler(name)}
+                  >
+                    {renderComponent === 'person' && (
```
ctizen, line 79: Why do we need such specific conditions inside?

packages/web/src/components/ui/SidebarCategorySearch/SidebarCategorySearch.tsx
```diff
@@ -0,0 +1,104 @@
...
+            <Icon kind='cross' color='#8E8AA0' className={styles.clear} onClick={clearInput} />
+          </span>
+        )}
+      </div>
+
+      <div className={styles.groups}>
+        {list?.length > 0 &&
+          list.map(({ group, children }) => (
+            <div key={group}>
+              {list.length > 1 && <div className={styles.groupName}>{group}</div>}
```
ctizen, line 68: What's going on here? Looks like the input data format is messy, or vars naming, or both.

packages/web/src/components/ui/Card/Card.tsx
```diff
@@ -0,0 +1,80 @@
...
+  description,
+  packageTags,
+  featuredSites,
+  vulnerablePackage,
+  variant,
+}: CardProps) {
+  return (
+    <div className={clsx(styles.card, variant && styles[variant])}>
+      <div className={styles.cardTop}>
+        <header className={styles.header}>
```
spalt08, line 42: I was wondering if it's worth using a header tag per each card? My concern is that may affect SEO some how

https://www.w3schools.com/tags/tag_header.asp

KnoWabLe: [W3 4.3.8 The header element](https://html.spec.whatwg.org/#the-header-element)

> A [header](https://html.spec.whatwg.org/#the-header-element) element is intended to usually contain a heading (an [h1](https://html.spec.whatwg.org/#the-h1,-h2,-h3,-h4,-h5,-and-h6-elements)–[h6](https://html.spec.whatwg.org/#the-h1,-h2,-h3,-h4,-h5,-and-h6-elements) element or an [hgroup](https://html.spec.whatwg.org/#the-hgroup-element) element), but this is not required. The [header](https://html.spec.whatwg.org/#the-header-element) element can also be used to wrap a section's table of contents, a search form, or any relevant logos.

Probably you're right, if it is not within `<section>` tag, it is not worth it.

packages/web/src/components/ui/Chip/Chip.module.scss
```diff
@@ -0,0 +1,127 @@
...
+
+.sans-serif {
+  font-family: $font-sans-serif;
+}
+
+.monospace {
+  font-family: $font-monospace;
+}
+
+.400 {
```
spalt08, line 96: I'm not sure if the selector name is valid since my linter shows an issue on this line

PR #53, opened 2022-09-02
feat: Data layer and flow rework

zardak:


packages/shared/src/database/entities/webPageScan.ts
```diff
@@ -0,0 +1,55 @@
...
+    Processed = 'processed',
+    Unsupported = 'unsupported',
+    Protected = 'protected',
+    Failed = 'failed',
+  }
+}
+
+export type ScanResultPackage = DetectedPackage;
+
+export type ScanResult = {
```
spalt08, line 53: Why just not
```ts
export namespace WebPageScan {
  // export enum Status ...
  
  export type Result = {
    packages: DetectedPackage[];
  }
}
```

packages/shared/src/internalApi/api.ts
```diff
@@ -1,23 +1,23 @@
 import fetch, { RequestInit } from 'node-fetch';
-import { getInternalApiOrigin } from '../utils/env';
+import { getGradeJsApiKey, getInternalApiRootUrl } from '../utils/env';
 
 export type DetectedPackage = {
   name: string;
-  possibleVersions: string[];
+  versionSet: string[];
   versionRange: string;
```
spalt08, line 7: It seems like we don't need `versionRange` anymore

packages/shared/src/internalApi/api.ts
```diff
@@ -1,23 +1,23 @@
...
   url: string;
-  status: WebsiteStatus;
+  status: WebPageScanStatus;
   detectedPackages: DetectedPackage[];
   updatedAt: string;
   createdAt: string;
 }
 
-export enum WebsiteStatus {
+export enum WebPageScanStatus {
```
spalt08, line 20:
```ts
export namespace WebPageScan {
  export enum Status {
    // ..
  }
}

**Review by spalt08 — CHANGES_REQUESTED** [review]

File: packages/public-api/src/router.ts [hunk 4]
```diff
@@ -2,64 +2,58 @@ import * as trpc from '@trpc/server';
...
+  export type PackageMetadataResponse = SerializableEntity<PackageMetadata>;
 }
 
 export const appRouter = trpc
   .router<Context>()
-  .mutation('syncWebsite', {
-    input: z.string().regex(hostnameRe),
-    async resolve({ input: hostname }) {
-      const webpages = await getWebPagesByHostname(hostname);
+  .mutation('requestWebPageScan', {
```
spalt08, line 33: Why have you removed `syncWebsite` mutation?

packages/public-api/src/projections/syncPackageUsageByHost.ts
```diff
@@ -0,0 +1,46 @@
...
+    throw new Error('Scan was not completed');
+  }
+
+  const packageUsageByHostRepo = em.getRepository(PackageUsageByHostnameProjection);
+  const webPageScanRepo = em.getRepository(WebPageScan);
+  const webPageRepo = em.getRepository(WebPage);
+
+  const previousScan = await webPageScanRepo
+    .createQueryBuilder('webpagescan')
+    .where('webpagescan.id < :newId', { newId: newScan.id })
```
spalt08, line 15: I was wondering if you need `where webpagescan.path === path` here?

packages/public-api/src/projections/syncPackageUsageByHost.ts
```diff
@@ -0,0 +1,46 @@
...
+  const packageUsageEntities = newScan.scanResult.packages.map((sourcePackage) =>
+    packageUsageByHostRepo.create({
+      hostname: relatedWebPage.hostname,
+      sourceScan: newScan,
+      packageName: sourcePackage.name,
+      packageVersionSet: sourcePackage.versionSet,
+    })
+  );
+
+  if (packageUsageEntities) {
```
spalt08, line 43: `packageUsageEntities.length > 0` ?

packages/public-api/src/website/service.ts
```diff
@@ -1,80 +1,116 @@
...
+  let webPageEntity = await webPageRepo
+    .createQueryBuilder('web_page')
+    .where('web_page.hostname_id = :hostnameId', { hostnameId: hostnameEntity.id })
+    .andWhere('web_page.path = :path', { path: url.pathname })
+    .limit(1)
+    .getOne();
+  if (!webPageEntity) {
+    webPageEntity = await webPageRepo.save({
+      hostname: hostnameEntity,
+      path: url.pathname,
```
spalt08, line 37: what about query params or URL hashes?

packages/shared/src/database/migrations/1662077882511-DataLayerRework.ts
```diff
@@ -0,0 +1,142 @@
...
+        "id" serial primary key,
+        "source_scan_id" integer not null references "web_page_scan"("id"),
+        "vulnerabilities" jsonb not null,
+        "created_at" timestamp,
+      );
+      
+      CREATE INDEX "scans_with_vulnerabilities_projection_created_at" ON "scans_with_vulnerabilities_projection" ("created_at" DESC);
+    `);
+
+    await queryRunner.query(`
```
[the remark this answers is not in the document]
zardak: Good catch, fixed

packages/shared/src/internalApi/api.ts
zardak: Keeping it for backwards-compatibility for now, until we rework the scan data structure itself

packages/public-api/src/website/service.ts
zardak: Open for discussion, but I can hardly imagine a case when different bundles would be served based on query params, let alone the hash, which is solely a client-side URL part.

packages/shared/src/utils/env.ts
```diff
@@ -32,7 +33,11 @@ export const isStaging = () => getNodeEnv() === 'staging';
 export const isDevelopment = () => getNodeEnv() === 'development';
 export const isTest = () => getNodeEnv() === 'test';
 
-export const getInternalApiOrigin = () => getEnv(Env.InternalApiOrigin);
+// TODO: DROP DEFAULTS
```
ctizen, line 36: Looks like it can be done already

packages/shared/src/database/migrations/1662077882511-DataLayerRework.ts
```diff
@@ -0,0 +1,131 @@
...
+      );
+      
+      CREATE INDEX "web_page_scan_web_page_id_created_at" ON "web_page_scan" ("web_page_id", "created_at" DESC);
+    `);
+
+    await queryRunner.query(`
+      CREATE TABLE "package_usage_by_hostname_projection" (
+        "id" serial primary key,
+        "hostname_id" integer not null references "hostname"("id"),
+        "source_scan_id" integer not null references "web_page_scan"("id"),
```
ctizen, line 50: I'd recommend against using the explicit `references` clause, as it may cause problems more than it solves. I believe that in most cases conventional references are enough (e.g. on the application level).

packages/public-api/src/website/service.ts
```diff
@@ -1,80 +1,120 @@
...
+    });
   }
+
+  return webPageEntity;
 }
 
-export async function getWebPagesByHostname(hostname: string) {
-  return getRepository(WebPage).find({ hostname });
+export async function requestWebPageScan(url: string) {
```
spalt08, line 44: According to the naming I wouldn't expect that function returns a cached scan. Can we rename it to something like `getOrRequestWebPageScan`?

packages/shared/src/database/migrations/1662077882511-DataLayerRework.ts
```diff
@@ -0,0 +1,131 @@
...
+         (SELECT jsonb_agg(r)
+           FROM (
+             SELECT "package_version", count("package_version")
+             FROM
+               "package_usage_by_hostname_projection" as "package_usage_subquery",
+               jsonb_array_elements_text("package_usage_subquery"."package_version_set") AS "package_version"
+             WHERE "package_usage_subquery"."package_name" = "package_usage"."package_name"
+             GROUP BY "package_version"
+           ) as r
+         ) as json_data
```
spalt08, line 83: Shouldn't it be `version_popularity` instaed of `json_data`?

packages/shared/src/database/migrations/1662077882511-DataLayerRework.ts
```diff
@@ -0,0 +1,131 @@
...
+      );
+      
+      CREATE INDEX "scans_with_vulnerabilities_projection_created_at" ON "scans_with_vulnerabilities_projection" ("created_at" DESC);
+    `);
+
+    await queryRunner.query(`
+      CREATE MATERIALIZED VIEW "package_popularity_view" AS
+        SELECT
+        "package_usage"."package_name" AS "package_name",
+         count(DISTINCT "package_usage"."hostname_id") as "popularity_rank",
```
spalt08, line 73: I'd suggest to use `hostnames_total` instead of `popularity_rank`

packages/shared/src/database/migrations/1662077882511-DataLayerRework.ts
zardak: Could you please elaborate? I've never encountered problems with FKs, they are integral part of any relational DB.

packages/public-api/src/systemApiRouter.test.ts
```diff
@@ -0,0 +1,63 @@
...
+            versionRange: '17.0.0',
+            approximateByteSize: null,
+          },
+        ],
+      },
+    };
+
+    const sencWebPageScanResultMock = jest.spyOn(WebsiteService, 'syncWebPageScanResult');
+    sencWebPageScanResultMock.mockImplementation(async () => {
+      return getRepository(WebPageScan).create({
```
spalt08, line 54: nit: I am not following why you need to create an entity with a single `status` field. 
```
sencWebPageScanResultMock.mockImplementation(() => Promise.resolve({ status: WebPageScan.Status.Processed }))
```

packages/public-api/src/website/service.ts
```diff
@@ -1,80 +1,120 @@
...
-    where: { hostname },
+export async function syncWebPageScanResult(scanReport: SystemApi.ScanReport) {
+  const db = await getDatabaseConnection();
+
+  return await db.transaction(async (em) => {
+    const webPageScanRepo = em.getRepository(WebPageScan);
+
+    let scanEntity: WebPageScan;
+    if (scanReport.id) {
+      scanEntity = await webPageScanRepo.findOneOrFail({ id: parseInt(scanReport.id, 10) });
```
spalt08, line 86: Can you please re-check typeorm type for `bigint`? From my experience it is always represented as a `string`.

packages/shared/src/systemApi/api.ts
```diff
@@ -67,27 +56,37 @@ export async function fetchEndpoint<T>(
...
+    requestInit.headers = { 'Content-Type': 'application/json', 'X-Api-Key': getGradeJsApiKey() };
     requestInit.body = JSON.stringify(data);
   } else if (method === 'GET' && data) {
     for (const key of Object.keys(data)) {
       requestUrl.searchParams.append(key, `${data[key]}`);
     }
   }
 
-  // console.log('Request to internal API: ', requestUrl.toString(), requestInit);
+  console.log(requestUrl.toString());
```
spalt08, line 71: We probably need to comment/remove this line

packages/shared/src/systemApi/api.ts
```diff
@@ -67,27 +56,37 @@ export async function fetchEndpoint<T>(
...
 
   return fetch(requestUrl.toString(), requestInit)
-    .then((response) => response.json())
+    .then((response) => {
+      console.log(requestUrl.toString(), response.status);
```
spalt08, line 75: Same here

packages/shared/src/systemApi/api.ts
```diff
@@ -67,27 +56,37 @@ export async function fetchEndpoint<T>(
...
+      }
+
+      return { data: {} };
+    })
     .then((json: any) => {
       if (!json.data) {
         throw new Error('Invalid response format');
       }
 
+      console.log(requestUrl.toString(), json);
```
spalt08, line 88: And here

packages/web/src/store/slices/home.ts
```diff
@@ -3,7 +3,11 @@ import { client } from '../../services/apiClient';
 import { RootState } from '../';
 
 const parseWebsite = createAsyncThunk('home/submitWebsite', async (url: string) => {
-  await client.mutation('requestParseWebsite', url);
+  if (!url.startsWith('http://') && !url.startsWith('https://')) {
```
spalt08, line 6: nit: the URL validation and formatting should probably be inside a component logic

spalt08 (approved):
Left some minor comments

ctizen (approved):

PR #55, opened 2022-09-05
chore: get more data from npm registry

ctizen:
Fixes https://github.com/fingerprintjs/gradejs-backend/issues/179


packages/worker/src/npmRegistry/api.ts
```diff
@@ -55,14 +68,33 @@ export async function fetchPackageMetadata(name: string) {
...
 
   return {
-    description: document.description ?? undefined,
-    license: document.license ?? undefined,
+    description: document.description,
+    fullDescription: document.readme,
+    maintainers: (document.maintainers ?? []).map((author) => ({
+      name: author.name,
+      email: author.email,
+      avatar: `https://s.gravatar.com/avatar/${createHash('md5')
```
spalt08, line 76: I'd extract this into separate function

spalt08 (changes requested):

spalt08:
Can you please also update `PackageMetadata` DB model?

packages/worker/src/tasks/syncPackageIndexBatch.ts
```diff
@@ -31,9 +31,13 @@ async function syncPackage(name: string) {
         name: name,
         latestVersion: statsAndMetadata.latestVersion,
         description: statsAndMetadata.description,
+        fullDescription: statsAndMetadata.fullDescription,
+        maintainers: statsAndMetadata.maintainers,
+        tags: statsAndMetadata.tags,
```
spalt08, line 36: nit: I was thinking about keeping the `keywords` naming instead of `tags`. What's your take?

packages/worker/src/npmRegistry/api.ts
```diff
@@ -55,14 +68,33 @@ export async function fetchPackageMetadata(name: string) {
...
+    })),
+    tags: document.keywords ?? [],
+    license: document.license,
     homepageUrl,
     repositoryUrl,
     updateSeq: Number(document._local_seq),
     updatedAt: new Date(document.time.modified),
     latestVersion: document['dist-tags'].latest,
     versionList,
+    versionSpecificValues: versionList.reduce((acc, el) => {
```
spalt08, line 88: nit: what if we make a single `versions: Record<string, VersionData>` property? So the `versionList` can be easily obtained by `Object.keys(versions)`

spalt08: Or `versionMap`

spalt08 (approved):

zardak (approved):

PR #59, opened 2022-09-07
feat: add react helmet for seo tags

ctizen:


packages/web/src/components/App.tsx
```diff
@@ -18,10 +19,25 @@ export function App({ locationChangeHandler }: AppProps) {
...
-      <Route path='*' element={<Navigate replace to='/' />} />
-    </Routes>
+    <>
+      <Helmet>
+        <title>Production Webpack Bundle Analyzer - GradeJS</title>
+        <meta
+          name='description'
+          content='GradeJS analyzes production JavaScript files and matches bundled NPM packages with specific version precision.'
+        />
+        <meta property='og:title' content='Production Webpack Bundle Analyzer - GradeJS' />
```
spalt08, line 29: Seems like `og:url` is also required. Can you please add it?
image

packages/web/src/components/Layout.tsx
```diff
@@ -0,0 +1,69 @@
...
+      <head>
+        {head.title.toComponent()}
+        {head.meta.toComponent()}
+        <meta charSet='utf-8' />
+        <meta name='viewport' content='width=device-width, initial-scale=1' />
+        <meta name='mobile-web-app-capable' content='yes' />
+        {/* TODO: preload all fonts after redesign
+          <link rel="preload" href={} as="font" type="font/woff2"/>
+        */}
+        <link rel='icon' type='image/svg+xml' href={'TODO'} />
```
spalt08, line 44: I guess we should comment `TODO` hrefs here as well

packages/web/src/components/pages/WebsiteResults.tsx
```diff
@@ -84,15 +86,30 @@ export function WebsiteResultsPage() {
...
-      host={hostname ?? ''}
-      vulnerabilities={vulnerabilities}
-      onFiltersApply={setFilters}
-    />
+    <>
+      <Helmet>
+        <title>{title}</title>
+        <meta name='description' content={description} />
+        <meta property='og:title' content={title} />
+        <meta property='og:description' content={description} />
```
spalt08, line 102: `og:url` is missing here as well

packages/web/src/server.tsx
```diff
@@ -25,6 +27,28 @@ app.get('/robots.txt', (_, res) =>
   })
 );
 
+function getScripts(statsStr: string) {
+  let stats: Record<string, string[]>;
+  const assets: { js: string[]; css: string[] } = { js: [], css: [] };
+  try {
+    stats = JSON.parse(statsStr).assetsByChunkName as Record<string, string[]>;
```
spalt08, line 34: I was wondering if we can use `webpack` typings for additional type guarding in case something changes inside webpack

spalt08 (changes requested):

packages/web/src/server.tsx
ctizen: Probably not (or at least I would not count on it). I think if something changes inside webpack we'll have much more interesting things other than updating stats format.

packages/web/src/index.tsx
```diff
@@ -12,7 +12,7 @@ const locationChangeHandler = initAnalytics();
 ReactDOM.hydrate(
   <Provider store={store}>
     <BrowserRouter>
-      <App locationChangeHandler={locationChangeHandler} />
+      <App locationChangeHandler={locationChangeHandler} hostname={window.location.host} />
```
spalt08, line 15: Why do you need to pass a property here if `location` is available with both `window` or `useLocation` hook inside any component?

packages/web/src/components/pages/WebsiteResults.tsx
spalt08: Still missing

ctizen: It's not. og:url tag is taken from upper scope with proper location.

packages/web/src/index.tsx
ctizen: We can't use `window` in server-rendered components. `useLocation` does not have enough data to make full link.

packages/web/src/components/App.tsx
```diff
@@ -14,14 +16,32 @@ type AppProps = {
  *   <App />
  * </BrowserRouter>
  */
-export function App({ locationChangeHandler }: AppProps) {
+export function App({ locationChangeHandler, hostname }: AppProps) {
```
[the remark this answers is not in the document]
ctizen: Not sure if env is a proper place for this. Maybe worth passing it with react context, sound ok?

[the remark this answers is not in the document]
ctizen: Fixed, assumed setting PUBLIC_ROOT_URL as env parameter

spalt08 (approved):

zardak (approved):

PR #60, opened 2022-09-08
wip: add loading bar, skeletons, mobile filters and error page

KnoWabLe:
image


spalt08 (approved):

packages/web/src/components/layouts/SearchResults/SearchResults.tsx
```diff
@@ -13,8 +13,32 @@ import CardList from '../../ui/CardList/CardList';
...
+  const [loading, setLoading] = useState<boolean>(pageLoading);
+  const loadingRef = useRef<LoadingBarRef>(null);
+
+  // FIXME: just for demo purposes to show how loading bar works
+  // Documentation: https://github.com/klendi/react-top-loading-bar
+  // Starts the loading indicator with a random starting value between 20-30 (or startingValue),
+  // then repetitively after an refreshRate (in milliseconds), increases it by a random value
+  // between 2-10. This continues until it reaches 90% of the indicator's width.
+  useEffect(() => {
+    loadingRef?.current?.continuousStart(10, 5000);
```
zardak, line 33: nit: We'll likely want a redux middleware which controls this based on async thunks and a separate top-level loading bar component.

packages/web/src/components/layouts/Error/Error.tsx
```diff
@@ -1,46 +1,161 @@
...
+          </div>
+
+          <div className={styles.searchWrapper}>
+            <SearchBar size='large' placeholder={actionTitle} />
+          </div>
+        </section>
+
+        <CardGroups>
+          <CardGroup title='But we have'>
+            <CardList cards={similarCards} />
```
zardak, line 149: nit: Trying to fit separate domain entities within a single component seems like burden. Feels like these `<CardList/>`'s should be separate components.

packages/web/src/components/layouts/SearchResults/SearchResults.tsx
```diff
@@ -183,69 +241,45 @@ export default function SearchResults() {
...
+              <SidebarMeta meta={metaItems} loading={loading} />
             </div>
 
             <div className={styles.sidebarItem}>
               <SidebarCategory
+                categoryName='Keywords'
                 keywordsList={keyWords}
                 selectedKeywords={selectedKeywords}
                 selectHandler={handleKeywordsChange}
                 renderComponent='chip'
```
zardak, line 258: nit: What about splitting these off into separate components and passing them back here through `children` or a render prop of sorts?

packages/web/src/components/ui/CardList/CardList.tsx
```diff
@@ -2,18 +2,26 @@ import React from 'react';
...
 };
 
-export default function CardList({ cards, variant = 'default' }: Props) {
+export default function CardList({ cards, variant = 'default', loading }: Props) {
   return (
     <div className={clsx(styles.grid, styles[variant])}>
-      {cards.map((card) => (
-        <Card key={card.id} {...card} />
-      ))}
+      {loading
```
zardak, line 17: @spalt08 @ctizen need your opinion on this one. I tend to favor splitting off skeletons for atom/molecule level components off into a separate component and having loading states only for organism-level components. What direction would we want to go with this one?

packages/web/src/components/ui/ChipGroup/ChipGroup.module.scss
```diff
@@ -20,3 +20,6 @@
     margin-right: 8px;
   }
 }
+
+.chipSkeleton {
```
zardak, line 24: Does this serve any purpose?

packages/web/src/components/ui/Header/Header.tsx
```diff
@@ -6,11 +6,15 @@ import { Icon } from '../Icon/Icon';
 import { trackCustomEvent } from '../../../services/analytics';
 
 type Props = {
-  variant?: 'default' | 'homepage';
+  variant?: 'default' | 'homepage' | 'error';
```
zardak, line 9: nit: Seems like an attempt to fit too much into a single component

packages/web/src/components/ui/PackagePreview/PackagePreview.tsx
```diff
@@ -139,38 +206,66 @@ export default function PackagePreview({ name, version, opened }: Props) {
...
+                    </Skeleton>
+                    <Skeleton>
+                      <div className={styles.statSubtitle}>out of 12 842</div>
+                    </Skeleton>
+                  </>
+                ) : (
+                  <>
+                    <div className={styles.statTitle}>
+                      385
+                      {/* or: <div className={clsx(styles.statRating, styles.statRatingRed)}> */}
```
zardak, line 244: What about adding a `rankingDelta` prop and deciding on class name based on number's sign?

packages/web/src/components/ui/PackagePreview/PackagePreview.tsx
```diff
@@ -182,75 +277,47 @@ export default function PackagePreview({ name, version, opened }: Props) {
...
-                  <div className={styles.popularityItem}>
-                    <div className={styles.popularityFill} style={{ height: '100%' }}>
-                      89 912
+                {modules.map(({ fill, uses, moduleVersion, bug }) => (
+                  <div className={styles.popularityItemWrapper}>
+                    <div className={styles.popularityItem}>
+                      {packageDetailsLoading ? (
+                        <div
+                          className={clsx(styles.popularityFill, styles.popularityFillSkeleton)}
+                          style={{ height: fill }}
```
zardak, line 286: We should be on lookout for these deoptimizations, this should definitely be a component / top-level const.

packages/web/src/components/ui/SitesList/SitesList.tsx
```diff
@@ -28,13 +44,13 @@ function Site({ image, name, packagesCount }: Site) {
...
 
-export default function SitesList({ sites, className }: Props) {
+export default function SitesList({ sites, className, loading }: Props) {
   return (
     <div className={styles.sitesListWrapper}>
       <div className={clsx(styles.sitesList, className)}>
-        {sites.map((site) => (
-          <Site key={site.id} {...site} />
-        ))}
+        {loading
```
zardak, line 51: The `repeat` fn added would likely be a good fit here too.

packages/web/src/utils/helpers.tsx
```diff
@@ -1,3 +1,9 @@
+import React from 'react';
+
 export function formatNumber(x: number) {
   return x.toLocaleString();
 }
+
+export function repeat(times: number, children: React.ReactNode) {
+  return [...Array(times)].map((_, idx) => <React.Fragment key={idx}>{children}</React.Fragment>);
```
zardak, line 8: Please refrain from calling constructors without a `new` keyword. Also, spread can be replaced with `.fill` as in 
`new Array(times).fill(undefined).map(...)`.

packages/web/src/components/ui/CardList/CardList.tsx
spalt08: I also tend to split off skeletons, since the code of actual components looks much cleaner

ctizen: I see no evil in both options at the moment, as the suggestion doesn't change much in total. E.g. splitting skeletons from here will lead to style duplication somewhere else and thus make it less clean. The complexity is either here or there, we can't just ignore it.

packages/web/src/components/layouts/SearchResults/SearchResults.tsx
ctizen: Agree. LoadingBar is likely to reside in the top-level app component.

packages/web/src/components/ui/ChipGroup/ChipGroup.tsx
```diff
@@ -16,15 +19,23 @@ export default function ChipGroup({
...
-        {chips.map((chip) => (
-          <Chip key={chip} className={styles.chip} size={size} font={font} fontSize={fontSize}>
-            {chip}
-          </Chip>
-        ))}
+        {loading
+          ? repeat(
+              4,
+              <Skeleton variant='rounded' className={styles.chip}>
+                <Chip>Lorem ipsum</Chip>
```
ctizen, line 31: Are we gonna render skeleton elements in SSR? If so, I'd like to not have any dummy text inside skeletons. If not, we probably must ensure that the dummy text will not accidentally pass into generated html.

packages/web/src/components/ui/Header/Header.tsx
KnoWabLe: How would you do it otherwise? Their purpose here is just to slightly modify styles (spacing and which elements to show in case of error page) to match design differences between places where they are used. Is that really to much for one component/wrapper to handle?

packages/web/src/components/ui/PackagePreview/PackagePreview.tsx
```diff
@@ -139,38 +206,66 @@ export default function PackagePreview({ name, version, opened }: Props) {
...
+                    <div className={styles.statSubtitle}>out of 12 842</div>
+                  </>
+                )}
               </div>
 
               <div className={clsx(styles.stat, styles.statListItemLarge)}>
                 <div className={styles.statHeader}>
-                  <Icon kind='dependency' color='#8E8AA0' className={styles.statIcon} />4 Dependency
+                  <Icon kind='dependency' color='#8E8AA0' className={styles.statIcon} />
+                  {!packageDetailsLoading && 4} Dependency
```
ctizen, line 263: Looks weird. Maybe hide/skeletonize the whole header?

KnoWabLe: Agreed, but according to design you already can see site address and its favicon/logo while loading. Package number just doesn't show in skeleton, it's just shows up eventually when loading is completed:

image

Maybe it's a design mistake and we should add skeleton for packages as well?

ctizen: Looks like a design mistake indeed. Ok, I'll write it down on my design issues list.

packages/web/src/components/ui/CardList/CardList.tsx
zardak: @KnoWabLe what's your opinion on restructuring skeletons like follows
```
/Component
  Component.scss
  Component.tsx
  ComponentSkeleton.tsx
```

packages/web/src/components/ui/Header/Header.tsx
zardak: This relates more to correctly segregating the responsibilities, rather than amount of code or variants.

I.e.:
* Passing links to render as a render prop. This gets rid of `display: none;` hack
* The differing padding between variants might be solved by underlying content (or, rather, the parent component)

packages/web/src/components/layouts/SearchResults/SearchResults.tsx
KnoWabLe: I propose leaving it as is for demo purposes, before actual data could be used. Maybe we could replace this component with our own solution or perhaps use more redux-compatible library/package.

packages/web/src/components/layouts/Error/Error.tsx
KnoWabLe: Added as `TODO` for future task/refactor.

packages/web/src/components/ui/PackagePreview/PackagePreview.tsx
KnoWabLe: Added as `TODO` for future `PackagePreview` refactor.

packages/web/src/components/ui/PackagePreview/PackagePreview.tsx
KnoWabLe: Added as `TODO` for future `PackagePreview` refactor.

zardak (approved):

PR #62, opened 2022-09-14
feat: extendable scan report format

spalt08:
Frontend part is broken


packages/worker/src/tasks/syncPackageIndexBatch.ts
```diff
@@ -1,7 +1,7 @@
-import { internalApi, PackageMetadata } from '@gradejs-public/shared';
+import { systemApi, SystemAPI, PackageMetadata } from '@gradejs-public/shared';
```
ctizen, line 1: Can we use the same identifier for types and runtime? Having two identifiers with different casing looks misleading.

packages/public-api/src/website/service.ts
```diff
@@ -106,15 +112,12 @@ export async function syncWebPageScanResult(scanReport: SystemApi.ScanReport) {
   });
 }
 
-function mapInternalWebsiteStatus(status: internalApi.WebPageScan.Status) {
+// TODO: add protected website use-case
```
zardak, line 115: Do you intentionally want to merge this and fix after?

zardak (approved):

packages/public-api/src/website/service.ts
spalt08: Yes

spalt08: Task: https://github.com/fingerprintjs/gradejs-backend/issues/195

packages/worker/src/tasks/syncPackageIndexBatch.ts
spalt08: fixed f6de5fb

ctizen (approved):

PR #64, opened 2022-09-22
Feature/markup integration search results

ctizen:


packages/public-api/src/clientApiRouter.ts
```diff
@@ -50,10 +53,19 @@ type RequestWebPageScanResponse = Pick<WebPageScan, 'status' | 'finishedAt'> & {
...
+      return await requestWebPageRescan(url);
+    },
+  })
+  .query('getWebPageScan', {
     input: z.string().url(),
     async resolve({ input: url }) {
-      const scan = await getOrRequestWebPageScan(url);
+      const scan = await getWebPageScan(url);
+      if (!scan) {
+        return { id: '', status: WebPageStatusNoData.NoData };
```
zardak, line 67: Do we need to reinvent a 404 here?

packages/web/src/components/pages/WebsiteResults.tsx
```diff
@@ -87,13 +111,16 @@ export function WebsiteResultsPage() {
...
-        packages={packagesFiltered ?? []}
-        host={hostname ?? ''}
+        searchQuery={address ?? ''}
+        packages={packagesFiltered}
+        packagesStats={packagesStats}
         vulnerabilities={vulnerabilities ?? {}}
-        onFiltersApply={setFilters}
+        vulnerabilitiesCount={vulnerabilitiesCount}
+        keywordsList={keywordsList}
+        status={{ status, lastScanDate }}
```
zardak, line 123: Deopt

packages/public-api/src/website/service.ts
```diff
@@ -40,38 +40,41 @@ export async function findOrCreateWebPage(url: URL, em: EntityManager) {
...
-      .where('scan.web_page_id = :webPageId', { webPageId: webPageEntity.id })
-      .orderBy('scan.created_at', 'DESC')
-      .limit(1)
-      .getOne();
+export async function requestWebPageRescan(url: string) {
+  const parsedUrl = new URL(url);
+  const db = await getDatabaseConnection();
+  const lastScan = await getWebPageScan(url);
+  if (lastScan && Date.now() - lastScan?.createdAt.getTime() < RESCAN_TIMEOUT_MS) {
+    return false;
```
zardak, line 63: I'd still resort to simply returning an older scan if timeout hasn't yet passed. And had two fns like `getWebPageScan` and `getOrRequestWebPageScan`.

ctizen: We don't use any data returned from `requestWebPageRescan` and we don't call `getWebPageScan` subsequently, so I don't see any reason why it should return any data.

packages/public-api/src/clientApiRouter.ts
```diff
@@ -50,15 +55,24 @@ type RequestWebPageScanResponse = Pick<WebPageScan, 'status' | 'finishedAt'> & {
...
+      const scan = await getWebPageScan(url);
+      if (!scan) {
+        return null;
+      }
 
       const scanResponse: RequestWebPageScanResponse = {
         id: scan.id.toString(),
         status: scan.status,
-        finishedAt: scan.finishedAt,
+        finishedAt: formatDate(scan.finishedAt ?? new Date()),
```
spalt08, line 75: I don't get why do we need to have a default value for `finishedAt`?

packages/public-api/src/website/service.ts
spalt08: I'd suggest to use only `getWebPageScan` for now and make a separate `rescan` button that will trigger `requestWebPageRescan`

packages/public-api/src/clientApiRouter.ts
```diff
@@ -39,8 +37,15 @@ function mergeRegistryMetadata(
   }));
 }
 
-type RequestWebPageScanResponse = Pick<WebPageScan, 'status' | 'finishedAt'> & {
+enum WebPageStatusNoData {
+  NoData = 'noData',
```
zardak, line 41: Seems like this is no longer needed

packages/public-api/src/clientApiRouter.ts
```diff
@@ -50,15 +55,24 @@ type RequestWebPageScanResponse = Pick<WebPageScan, 'status' | 'finishedAt'> & {
...
+      return await requestWebPageRescan(url);
+    },
+  })
+  .query('getWebPageScan', {
     input: z.string().url(),
     async resolve({ input: url }) {
-      const scan = await getOrRequestWebPageScan(url);
+      const scan = await getWebPageScan(url);
+      if (!scan) {
+        return null;
```
zardak, line 69: What's the reasoning behind returning a `null` instead of throwing a `NotFound`?

packages/public-api/src/website/service.ts
zardak: We can merge this as-is, https://github.com/gradejs/gradejs/pull/65 covers the issue

packages/web/src/components/pages/Home.tsx
```diff
@@ -9,24 +9,30 @@ export function HomePage() {
...
     trackCustomEvent('HomePage', 'WebsiteSubmitted');
     // TODO: error state of input field, e.g. when empty
-    if (!address.startsWith('http://') && !address.startsWith('https://')) {
-      address = 'https://' + address;
-    }
     await dispatch(parseWebsite(address));
+
+    // TODO: properly handle history/routing
+    if (!state.isLoading && !state.isFailed && address) {
+      navigate(`/scan/${encodeURIComponent(address)}`);
```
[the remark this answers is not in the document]
zardak: Let's move the discussion to #65 and merge this as-is

PR #65, opened 2022-09-22
feature: redesign storage rework

zardak:


spalt08:
Storybook is broken
https://gradejs-git-feature-redesign-storage-rework-fp-pro.vercel.app/storybook/?path=/story/layouts-home--default

packages/web/src/components/containers/SearchBarContainer.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { useScanResult } from '../../store/hooks/useScanResult';
+import SearchBar from '../ui/SearchBar/SearchBar';
+import { useNavigate } from 'react-router-dom';
+
+type Props = {
+  size?: 'default' | 'large';
+  placeholder?: string;
+};
+
+// TODO: Dedupe logic at Home component
```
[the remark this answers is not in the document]
zardak: Both current and proposed approaches are equally bad IMO, we'll need to look into reworking the search bar to fit all the usecases later.

packages/web/src/store/hooks/useScanResult.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    try {
+      const prefixedScanUrl =
+        !scanUrl.startsWith('http://') && !scanUrl.startsWith('https://')
+          ? `https://${scanUrl}`
+          : scanUrl;
+
+      const parsedUrl = new URL(prefixedScanUrl);
+      const shortenedUrl = `${parsedUrl.hostname}${parsedUrl.pathname}`;
```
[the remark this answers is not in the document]
zardak: I fail to see why would this should be done now rather than during SEO pass

packages/web/src/components/pages/WebsiteResults.tsx
```diff
@@ -84,23 +57,14 @@ export function WebsiteResultsPage() {
...
-          trackCustomEvent('HostnamePage', 'ClickReport_Invalid');
-        }}
-        */
+        host={displayUrl ?? ''}
       />
     );
   }
 
-  const title = `List of NPM packages that are used on ${address} - GradeJS`;
+  const title = `List of NPM packages that are used on ${displayUrl} - GradeJS`;
```
spalt08, line 65: should be `hostname`, not the full URL.
(also in description)

packages/web/src/components/pages/Home.tsx
```diff
@@ -1,39 +1,28 @@
...
   const navigate = useNavigate();
-  const dispatch = useAppDispatch();
-  const state = useAppSelector(homeDefaultSelector);
-
-  const handleDetectStart = useCallback(async (address: string) => {
-    trackCustomEvent('HomePage', 'WebsiteSubmitted');
-    // TODO: error state of input field, e.g. when empty
-    await dispatch(parseWebsite(address));
+  const [requestedScanUrl, setRequestedScanUrl] = useState<string | undefined>(undefined);
+  const { displayUrl, scanResult } = useScanResult(requestedScanUrl, false, true);
```
spalt08, line 10: Can we use an option object here?

packages/web/src/store/hooks/scan/useScanResult.ts
```diff
@@ -0,0 +1,44 @@
...
+  useEffect(() => {
+    let timeoutId: number | undefined;
+    if (
+      pollWhilePending &&
+      normalizedUrl &&
+      scanResult?.scan?.status === 'pending' &&
+      !scanResult?.isLoading &&
+      !scanResult?.error
+    ) {
+      timeoutId = window.setTimeout(() => {
```
spalt08, line 35: Shouldn't it be `setInterval`?

spalt08 (approved):

packages/web/src/store/hooks/scan/useScanResult.ts
zardak: No, it's working as intended, as it's far easier to manage a `setTimeout` in `useEffect`, rather than a `setInterval`. Be mindful that `useEffect` is being triggered on each dependent value update, thus initiating a `setTimeout` each time a  scan is still in `pending` state and `isLoading` is no longer in effect.

PR #66, opened 2022-09-26
feat: add package page

KnoWabLe:


packages/web/src/components/layouts/PackagePage/PackagePage.tsx
```diff
@@ -0,0 +1,596 @@
...
+  // FIXME: just for demo purpose
+  useEffect(() => {
+    setTimeout(() => {
+      setLoading(false);
+    }, 5000);
+  }, []);
+
+  const sorts = ['weight', 'popularity', 'versions'];
+
+  const loadedData = {
```
ctizen, line 48: Can you please move it to mocks file?

packages/web/src/components/layouts/PackagePage/PackagePage.tsx
```diff
@@ -0,0 +1,596 @@
...
+                  type='button'
+                  className={clsx(
+                    styles.sortButtonMobile,
+                    sortField === name && styles.sortButtonMobileActive
+                  )}
+                  onClick={() => requestSort(name)}
+                >
+                  {name[0].toUpperCase() + name.slice(1)}
+
+                  {/* TODO: not sure how to handle it better right now by using <Icon> component */}
```
ctizen, line 131: You can use `currentColor` magic value in your svg, like `stroke='currentColor'` and then set color with `color` prop of `Icon`. See sprite/question-mark.svg for example.

packages/web/src/components/layouts/PackagePage/PackagePage.tsx
```diff
@@ -0,0 +1,596 @@
...
+        </div>
+      </Modal>
+
+      <StickyDefaultHeader showSearch />
+
+      <div className={styles.packagePage}>
+        <Container>
+          <div className={styles.packagePageGrid}>
+            <div className={styles.content}>
+              <h1>hoist-non-react-statics</h1>
```
ctizen, line 169: Some unexpected hardcode:)

packages/web/src/components/layouts/PackagePage/PackagePage.tsx
```diff
@@ -0,0 +1,596 @@
...
+                    </span>
+                  </div>
+                </div>
+
+                <div className={styles.packages}>
+                  {loading ? (
+                    repeat(5, <Skeleton width='100%' height={100} variant='rounded' />)
+                  ) : (
+                    <>
+                      <PackageVersion
```
ctizen, line 291: Could you please move the data from here into `loadedData` and update render logic?

packages/web/src/components/layouts/PackagePage/PackagePage.tsx
```diff
@@ -0,0 +1,596 @@
...
+                          <Skeleton className={styles.sidebarLinkTextSkeleton} />
+                        </span>
+                        <span className={styles.sidebarLinkSkeleton}>
+                          <Skeleton width={18} className={styles.sidebarLinkIconSkeleton} />
+                          <Skeleton className={styles.sidebarLinkTextSkeleton} />
+                        </span>
+                      </>
+                    ) : (
+                      <>
+                        <a
```
ctizen, line 384: Same here about moving data to `loadedData`. You can use `PackagePreview` component as a reference for input data format, as it's highly likely that format will be similar here.

packages/web/src/components/layouts/PackagePage/PackagePage.tsx
```diff
@@ -0,0 +1,596 @@
...
+                      </div>
+                      <Skeleton width='90%' />
+                      <Skeleton width='50%' />
+                    </>
+                  ) : (
+                    <>
+                      <div className={styles.sidebarItemTitle}>About</div>
+                      <p className={styles.sidebarItemText}>
+                        A JavaScript date library for parsing, validating, manipulating, and
+                        formatting dates
```
ctizen, line 442: And here too.

packages/web/src/components/layouts/PackagePage/PackagePage.tsx
```diff
@@ -0,0 +1,596 @@
...
+                            <img
+                              className={styles.statImage}
+                              src='https://via.placeholder.com/52'
+                              alt=''
+                            />
+                          </div>
+
+                          <div className={styles.statContent}>
+                            <span className={styles.statTitle}>Author</span>
+                            <span className={styles.statValue}>jdalton</span>
```
ctizen, line 484: And here.
Also, what about more than one author? I think what you did in `PackagePreview` with multiple avatars looks great, maybe we can use it here too.

ctizen (changes requested):

packages/web/src/components/layouts/PackagePage/PackagePage.tsx
KnoWabLe: I get the idea and I've used it before, but in this case problem we need to change color of individual `path` inside `svg` to highlight specific arrow as active with darker color and make second one gray:
image
image

By using `<Icon>` and passing `color` prop will make both arrows, represented as `path`'s, use that color, showing them both as black or gray depending on direction.

Example with `<Icon kind='order' color={sortDirection === 'asc' ? '#212121' : '#DDDCE3'} />`:
image
image

packages/web/src/components/ui/CircularProgress/CircularProgress.tsx
```diff
@@ -0,0 +1,57 @@
...
+import styles from './CircularProgress.module.scss';
+
+type Props = {
+  progress: number;
+  size?: number;
+  strokeWidth?: number;
+};
+
+const CircularProgress = ({ progress, size = 48, strokeWidth = 5 }: Props) => {
+  const sizeWithTrack = size + strokeWidth;
```
zardak, line 11: lines 11-16 could me moved inside the `useMemo` below

packages/web/src/components/ui/PackageVersion/PackageVersion.module.scss
```diff
@@ -0,0 +1,299 @@
...
+}
+
+.packageModules {
+  display: grid;
+  grid-gap: 16px;
+  grid-template-columns: minmax(0, 1fr);
+}
+
+.packageModule {
+  //display: grid;
```
zardak, line 242: Please leave a TODO along commented-out code

packages/web/src/components/ui/PackageVersion/PackageVersion.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+  <PackageVersion {...args} />
+);
+
+Default.args = {
+  version: '18.2.0',
+  updateDate: '2 month ago',
+  uses: 23987,
+  size: 160,
+  sizeUnit: 'B',
+  sizeUnitShorthand: 'Byte',
```
zardak, line 22: Values (or actual usage) of `Shorthand` and non-shorthand props are swapped

packages/web/src/components/ui/PackageVersion/PackageVersion.tsx
```diff
@@ -0,0 +1,165 @@
...
+  size: number;
+  sizeUnitShorthand: string;
+  sizeUnit: string;
+  modulesCount: number;
+  modules: Module[];
+  entries: EntryPoint[];
+  opened?: boolean;
+};
+
+const PackageVersion = ({
```
zardak, line 35: nit: `PackageVersionDetailsBlock` or something along the lines of. Current same suggests more about displaying a package version string like `1.0.0`

packages/web/src/components/ui/PackageVersion/PackageVersion.tsx
```diff
@@ -0,0 +1,165 @@
...
+  entry: string;
+};
+
+type Props = {
+  version: string;
+  updateDate: string;
+  uses: number;
+  size: number;
+  sizeUnitShorthand: string;
+  sizeUnit: string;
```
zardak, line 28: Also regarding these, I'd always resort to passing everything around in bytes and formatting in-place after. We should already have a function to format bytes to readable size laying around.

zardak (changes requested):

spalt08 (approved):

ctizen (approved):

PR #70, opened 2022-09-28
feat: add showcase block to home and error pages

zardak:


packages/public-api/src/clientApiRouter.ts
```diff
@@ -49,6 +50,43 @@ type RequestWebPageScanResponse = {
...
+            showcasedScan.scan.scanResult?.identifiedPackages.slice(0, 6).map((pkg) => pkg.name) ??
+            [],
+          totalCount: showcasedScan.scan.scanResult?.identifiedPackages.length ?? 0,
+        },
+      }));
+
+      const scansWithVulnerabilities = showcaseData.scansWithVulnerabilities.map(
+        (scanWithVulnerabilities) => ({
+          hostname: {
+            hostname: scanWithVulnerabilities.sourceScan.webPage.hostname.hostname,
```
ctizen, line 75: Why such weird nesting?

zardak: These are, effectively, entities. I expect to land additional fields like `favicon` here later. Anyways, this is non-persistent data structure and can be easily refactored later.

ctizen (approved):

packages/web/src/components/ui/Card/PackagesBySourceCard.tsx
```diff
@@ -17,7 +17,7 @@ const PackagesBySourceCard = ({
   packages,
   morePackagesCount,
 }: PackagesBySourceCardProps) => (
-  <Card title={sourceTitle} icon={sourceIcon}>
+  <Card title={sourceTitle} icon={sourceIcon} to={`/scan/${sourceTitle}`}>
```
spalt08, line 20: shouldn't it be `sourcePageUrl`?

spalt08 (approved):

packages/shared/src/database/migrations/1664181194347-ShowcasedScans.ts
```diff
@@ -0,0 +1,19 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class ShowcasedScans1664181194347 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+      create table "showcased_scan" (
+        "id" serial primary key,
+        "display_order" smallint not null default 0,
```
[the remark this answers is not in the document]
zardak: Until we have a clear-ish vision of what we're trying to accomplish with it in mind, I'd rather not add a column which will be null for 99.9% entries of the table.

packages/web/src/components/ui/Card/PackagesBySourceCard.tsx
zardak: Not sure what you meant, but I'll probably split the title and the url into separate props. This component originally served two purposes: displaying packages by source scan and displaying packages by author.

spalt08 (approved):

PR #71, opened 2022-09-28
Fix/tmp disable package dropdown

ctizen:


packages/web/src/components/ui/PackagePreview/PackagePreview.tsx
```diff
@@ -59,10 +59,11 @@ export default function PackagePreview({
               </span>
             )}
           </div>
-
+          {/*
```
zardak, line 62: Can we keep just this as a comment with a TODO and delete `open` altogether for now? This seems quite heavy to read and wouldn't be hard to restore after.

zardak (approved):

spalt08:
Can you please remove the details button?
image

zardak (approved):

PR #72, opened 2022-09-28
fix: make authors grouped avatars

ctizen:


packages/web/src/components/ui/Avatar/Avatar.tsx
```diff
@@ -1,15 +1,36 @@
 import React from 'react';
 import styles from './Avatar.module.scss';
+import stylesTooltip from '../Tooltip/Tooltip.module.scss'; // TODO: is it ok to reference external styles?
```
spalt08, line 3: I guess it may work as an ad-hoc solution, but I'd like to avoid it in the future

packages/web/src/components/ui/Avatar/Avatar.tsx
```diff
@@ -1,15 +1,36 @@
...
 };
 
-const Avatar = ({ src, alt = '' }: Props) => (
-  <div className={styles.avatarItem}>
-    <img src={src} className={styles.avatarImage} alt={alt} />
-  </div>
-);
+const tooltipOptions: Config = { placement: 'top-end', offset: [20, 13] };
+
+const Avatar = ({ src, altAsTooltip = false, alt = '' }: Props) => {
```
spalt08, line 14: Can you please also add a storybook for a component?

spalt08 (approved):

zardak (approved):

PR #78, opened 2022-09-30
feat: add package filtering

zardak:


spalt08:
image

packages/web/src/components/ui/ProblemsList/ProblemsList.tsx
```diff
@@ -1,22 +1,30 @@
 import React from 'react';
 import styles from './ProblemsList.module.scss';
 import Checkbox from '../Checkbox/Checkbox';
+import { capitalizeWord } from '../../../utils/helpers';
 
 type Props = {
   keywordsList: string[];
   selectedKeywords: string[];
-  selectHandler: (keyword: string) => void;
+  selectHandler: (newKeywords: string[]) => void;
```
spalt08, line 9: nit: `onChange`

packages/web/src/components/ui/SearchResultsSidebar/SearchResultsSidebar.tsx
```diff
@@ -18,117 +18,142 @@ import { Button } from '../index';
...
-          keywordsList={keyWords}
-          selectedKeywords={selectedKeywords}
-          selectHandler={handleKeywordsChange}
-          returnButton={() => setModalKeywordsOpen(false)}
-          resetGroup={() => setSelectedKeywords([])}
+          keywordsList={availableFilters.keywords}
+          selectedKeywords={selectedFilters.keywords}
+          selectHandler={handleKeywordsFilterChange}
+          returnButton={closeModalHandler}
+          resetGroup={handleKeywordsFilterReset}
```
spalt08, line 127: I guess it should be `onReturnClick` and `onResetClick`

packages/web/src/components/ui/SearchResultsSidebar/SearchResultsSidebar.tsx
```diff
@@ -18,117 +18,142 @@ import { Button } from '../index';
...
-            resetGroup={() => setSelectedProblems([])}
+            selectedKeywords={selectedFilters.traits}
+            returnButton={closeModalHandler}
+            resetGroup={handleTraitsFilterReset}
           >
             <ProblemsList
-              keywordsList={problems}
-              selectedKeywords={selectedProblems}
-              selectHandler={handleProblemsChange}
+              keywordsList={availableFilters.traits}
```
ctizen, line 141: Looks like renaming the 'problems' entity to 'traits' was unnecessary. However, my point is - we shouldn't use different terminology for the same entity, so I probably could agree with renaming to 'traits' _everywhere_, but now it's not the case.

packages/web/src/store/selectors/websiteResults.ts
```diff
@@ -52,7 +53,8 @@ const makeSelectScanPackagesByUrl = () =>
           !pkg.versionSet.some(
             (ver) => pkg.registryMetadata && semver.eq(pkg.registryMetadata.latestVersion, ver)
           ),
-        vulnerable: (scanData.vulnerabilities[pkg.name]?.length ?? 0) > 0,
+        vulnerable: (scanData.vulnerabilities[pkg.name]?.length ?? 0) > 0, // TODO: Drop
+        vulnerabilities: scanData.vulnerabilities[pkg.name] ?? [],
```
spalt08, line 57: I guess that may conflict with #80

packages/web/src/store/selectors/scanDisplayOptions/packageFilters.ts
```diff
@@ -0,0 +1,22 @@
+import { PackageTrait } from '../../slices/scanDisplayOptions';
+import { IdentifiedPackage } from '../websiteResults';
+
+export type OptimizedPackageFilters = {
```
ctizen, line 4: Are there non-optimized filters somewhere?

packages/web/src/store/selectors/scanDisplayOptions/scanDisplayOptions.ts
```diff
@@ -0,0 +1,59 @@
...
+};
+
+export const makeSelectScanDisplayOptions = () =>
+  createSelector(
+    [(state: RootState) => state.scanDisplayOptions, (_, scanUrl: string | undefined) => scanUrl],
+    (scanDisplayOptions, scanUrl = '') =>
+      scanDisplayOptions[scanUrl] ?? DEFAULT_SCAN_DISPLAY_OPTIONS
+  );
+
+export const makeSelectOptimizedScanDisplayOptions = () =>
```
spalt08, line 33: Why do you need to have both optimized and non-optimized scan display options? Can't we use sets everywhere?

packages/web/src/components/ui/KeywordsList/KeywordsList.tsx
```diff
@@ -7,17 +7,31 @@ import ChipGroup from '../ChipGroup/ChipGroup';
 type Props = {
   keywordsList: string[];
   selectedKeywords: string[];
-  selectHandler: (keyword: string) => void;
+  selectHandler: (newKeywords: string[]) => void;
```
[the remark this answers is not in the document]
zardak: Valid point, but components are in need of much heavier pass of refactoring.

packages/web/src/store/selectors/scanDisplayOptions/scanDisplayOptions.ts
zardak: That's in the plans for refactoring pass.

packages/web/src/components/ui/SearchResultsSidebar/SearchResultsSidebar.tsx
zardak: I see where you're coming from, but distinct names in business logic domain and UI presentation layer aren't a problem per se and encountered from time to time. In this very case I'd still resort to current naming having imminent additional features (like `duplicate`, etc.) in mind.

packages/web/src/store/selectors/scanDisplayOptions/packageFilters.ts
zardak: In the store itself, in raw serializable form.

packages/web/src/store/selectors/websiteResults.ts
zardak: Updated to use this field across the component.

packages/web/src/components/ui/SearchResultsSidebar/SearchResultsSidebar.tsx
ctizen: Different naming increases cognitive load and amounts of chaos in the codebase. It's not about logic domain or ui, it's about how you or someone else reads the code and understands its meaning and how different parts are connected to each other. I can approve this if you insist, though it is to be refactored in nearest future.

spalt08 (approved):

ctizen (approved):

packages/web/src/components/ui/SearchResultsSidebar/SearchResultsSidebar.tsx
zardak: The reasoning you gave was exactly the case why i chose to rename it the current way. "Package trait" or "Package feature" when talking about a "vulnerable" / "outdated" / "duplicate" package makes much more sense to me rather than a "Package problem". @spalt08 Your call.

PR #87, opened 2022-10-04
fix: hide script header

ctizen:


spalt08 (approved):

packages/web/src/components/ui/PackagePreview/PackagePreview.tsx
```diff
@@ -82,27 +82,29 @@ export default function PackagePreview({
       >
         <div className={styles.content}>
           <div className={styles.contentInner}>
-            <div className={styles.stat}>
-              <div className={styles.statHeader}>
-                <Icon kind='script' color='#8E8AA0' className={styles.statIcon} />
-                Script
+            {(pkg.containingScripts?.length ?? 0) > 0 && (
```
zardak, line 85: nit: We're usually sticking to this pattern across the codebase `!!pkg.containingScripts?.length`

zardak (approved):

packages/web/src/components/ui/PackagePreview/PackagePreview.tsx
```diff
@@ -82,27 +82,29 @@ export default function PackagePreview({
...
+              <div className={styles.stat}>
+                <div className={styles.statHeader}>
+                  <Icon kind='script' color='#8E8AA0' className={styles.statIcon} />
+                  Script
+                </div>
+                {detailsLoading ? (
+                  <ScriptSkeleton />
```
zardak, line 92: Also, seems like `<ScriptSkeleton>` will never be shown with current logic and data flow

packages/web/src/components/ui/PackagePreview/PackagePreview.tsx
spalt08: It should also work without `!!`

packages/web/src/components/ui/PackagePreview/PackagePreview.tsx
ctizen: True. I suggest dealing with it a bit later though.

PR #88, opened 2022-10-04
fix: add text overflow to chip text

KnoWabLe:
- add text overflow to chip text with max-width
- render icon span only if icon is present
- pass optional title for accessibility
- add few missing keys

Before:
image

After:
image

### DIFF
--- a/packages/web/src/components/ui/Chip/Chip.module.scss
+++ b/packages/web/src/components/ui/Chip/Chip.module.scss
@@ -2,6 +2,7 @@
 @import '~styles/responsive.scss';
 
 .chip {
+  max-width: 100%;
   cursor: pointer;
   display: inline-flex;
   align-items: center;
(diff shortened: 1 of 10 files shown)


packages/web/src/components/ui/Chip/Chip.module.scss
```diff
@@ -133,3 +133,18 @@
   font-size: 14px;
   line-height: 20px;
 }
+
+.text {
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  max-width: 258px;
```
spalt08, line 141: can we make chip `max-width: 100%` without any hard coded widths?

KnoWabLe:
Fixed a few more places where it can occur:
after-1
after-2
after-3

zardak (approved):

PR #97, opened 2022-10-06
fix: redirect from /w/ to /scan/

ctizen:


spalt08 (commented):
Please add www.*com -> *.com redirect

spalt08 (approved):

PR #100, opened 2022-10-07
fix: disallow robots on staging

ctizen:


packages/web/src/server.tsx
```diff
@@ -17,14 +17,18 @@ const staticDir = '/static';
...
 app.use(staticDir, express.static(path.join(__dirname, 'static')));
 app.get('/robots.txt', (_, res) =>
-  readFile(path.join(__dirname, '/robots.txt'), { encoding: 'utf-8' }, (err, data) => {
-    if (!err) {
-      res.send(data);
-    } else {
-      res.status(404);
-      res.send(null);
+  readFile(
+    path.join(__dirname, isStaging() ? '/robots.staging.txt' : '/robots.txt'),
```
spalt08, line 21: nit: can we extract paths (here and static dir) into variables for better readability?

spalt08 (approved):

PR #101, opened 2022-10-07
chore: tune page size

ctizen:


packages/web/src/components/ui/Hero/Hero.module.scss
```diff
@@ -3,7 +3,7 @@
 
 .hero {
   border-radius: 0 0 45px 45px;
-  background: url('~assets/hero-bg.png') 50% 0 no-repeat;
+  background: url('~assets/hero-bg.webp') 50% 0 no-repeat;
```
spalt08, line 6: Can we add an alias or a background color here? Some legacy browsers (such as safari) does not support webp

spalt08 (changes requested):

spalt08 (approved):

PR #104, opened 2022-10-10
feat: better logging using rollbar

ctizen:


packages/web/src/components/App.tsx
```diff
@@ -38,8 +39,22 @@ export function App({ locationChangeHandler }: AppProps) {
         <meta property='og:image' content='/static/sharing-image.png' />
       </Helmet>
       <Routes>
-        <Route index element={<HomePage />} />
-        <Route path='/scan/*' element={<WebsiteResultsPage />} />
+        <Route
+          index
+          element={
+            <RollbarContext context='/'>
```
spalt08, line 45: Can you use a single context per App component?

ctizen: This is intended to be used on the routes layer to make logs more specific, and it's not mandatory at all. I thought it would be better to have info about what route the error happened at.

zardak: Can't see why would we need this. Errors (should) clearly map to sources by stacktraces already.

packages/public-api/src/systemApiRouter.ts
```diff
@@ -9,8 +9,7 @@ systemApiRouter.post('/scan', async (req, res, next) => {
     const scanReport = systemApi.apiScanReportSchema.parse(req.body);
     await syncWebPageScanResult(scanReport);
   } catch (e) {
-    // TODO: report to rollbar
-    console.log(e);
+    logger.error(e);
```
zardak, line 12: This is a fall-through error handler, see the `next(e)` line. Not sure why would we explicitly need to track it here.

packages/web/src/server.tsx
```diff
@@ -54,31 +57,35 @@ function getScripts(statsStr: string) {
...
-  });
+      res.send(
+        '<!doctype html>' +
+          ReactDOMServer.renderToString(
+            <Layout js={js} css={css} head={helmet} env={getClientVars()} html={html} />
+          )
+      );
+    });
+  } catch (e) {
+    logger.error(e);
```
zardak, line 87: Did you intend to just swallow all of the errors here? IMO, we should still maintain proper 404 / 500 error codes.

zardak (changes requested):

packages/web/src/server.tsx
spalt08: I guess it was related to my hot fixes, since the `trpc` doesn't output errors in a console. I was wondering if we can make a single error handler for `trpc` like express error middleware

ctizen: This is not about api, it's about web. 
Added status(500) here. Probably we should pass the status code from inside components and set a proper status code when sending a response.

packages/web/src/components/App.tsx
ctizen: Ok, removed it

packages/public-api/src/middleware/common.ts
```diff
@@ -39,8 +39,7 @@ export function errorHandlerMiddleware(
     !(error instanceof UnauthorizedError) &&
     !(error instanceof ZodError)
   ) {
-    // TODO: add logger
-    console.error(error, req);
+    logger.error(error, req);
```
spalt08, line 42: Please make sure we log any 500 error without exposing a stack trace inside a response

spalt08 (approved):

zardak (approved):

ctizen:
Don't merge until env vars are set up on servers.

PR #105, opened 2022-10-12
feat: Initial SSR implementation

zardak:


packages/web/src/index.tsx
```diff
@@ -2,11 +2,18 @@ import React from 'react';
...
+import { createApplicationStore } from './store';
 import { App } from './components/App';
 import { initAnalytics } from './services/analytics';
 import 'styles/global.scss';
 
+declare global {
+  interface Window {
+    __INITIAL_STATE__?: object;
+  }
+}
```
ctizen, line 14: I think global.d.ts in web package root is a better place for this

packages/web/package.json
```diff
@@ -67,6 +67,7 @@
     "@types/lodash.memoize": "^4.1.7",
     "@types/react-helmet": "^6.1.5",
     "clsx": "^1.1.1",
+    "cross-fetch": "^3.1.5",
```
ctizen, line 70: There's already a `node-fetch` package in `shared`. I think we better use same packages across the monorepo.

packages/web/src/server.tsx
```diff
@@ -53,32 +56,79 @@ function getScripts(statsStr: string) {
...
     </Provider>
   );
+}
+
+async function renderTimeout(timeout: number, abort: AbortController) {
+  await setTimeout(timeout, null, { signal: abort.signal });
+  throw new Error('Timed out');
+}
+
+async function renderCycle(location: string, maxPasses = 3, passTimeout = 15000) {
```
ctizen, line 74: How are default values determined? I mean why it's 3 passes and not 2 or 5

ctizen (changes requested):

spalt08 (approved):

packages/web/src/index.tsx
zardak:
```
TS2669: Augmentations for the global scope can only be directly nested in external modules or ambient module declarations.
```

packages/web/package.json
zardak: `cross-fetch` is a maintained version of `isomorpgic-fetch`. They both use `node-fetch` internally, turning it into a polyfill. I'm rather unsure if we want to fiddle with globals ourselves instead of relying on a package.

packages/web/src/server.tsx
zardak: Determined by the requirements of the application. Effectively, this is a number longest chain of dependent requests (waterfall requests) required to render an application. Given that we've no dependent requests at the moment, two passes would be sufficient. I'll keep track and update this number as needed.

packages/web/src/index.tsx
ctizen: You just don't need the `declare global` wrapper there.

packages/web/package.json
ctizen: Ok, then may I ask you to replace node-fetch with cross-fetch in `packages/web/src/services/apiClient.ts` too?

zardak: Since it's a proper server polyfill, I'm dropping explicit references to `node-fetch` altogether.

spalt08 (approved):

PR #106, opened 2022-10-12
feat: add sort, search and 'not found' for results page

KnoWabLe:


spalt08:
Can you please also implement mobile sorting?

spalt08 (approved):

packages/web/src/components/pages/WebsiteResults.tsx
```diff
@@ -87,7 +114,26 @@ export function WebsiteResultsPage() {
...
+        setScanDisplayOptions({
+          scanUrl: normalizedUrl,
+          options: {
+            ...selectedDisplayOptions,
+            packageSorters: newSorters,
+          },
+        })
+      );
+    },
+    [dispatch, normalizedUrl, selectedDisplayOptions]
```
ctizen, line 136: The two new handlers look very much alike. Instead of these, it would probably be better to write some additional reducers in the scanDisplayOptions slice, which will allow avoiding overwriting the whole options object here.

packages/web/src/components/ui/SidebarCategory/SidebarCategorySearch.tsx
```diff
@@ -0,0 +1,56 @@
...
+  searchText: SearchText;
+  handleSearchTextChange: (newSearchText: SearchText) => void;
+};
+
+export default function SidebarCategorySearch({
+  categoryName,
+  searchText,
+  handleSearchTextChange,
+}: Props) {
+  const [searchValue, setSearchValue] = useState<string>('');
```
ctizen, line 20: Why do need the internal state here? The whole component looks like a simple wrapper.

packages/web/src/store/selectors/scanDisplayOptions/scanDisplayOptions.ts
```diff
@@ -46,14 +47,29 @@ export const makeSelectSortedAndFilteredScanPackages = () =>
...
     [makeSelectScanPackagesByUrl(), makeSelectOptimizedScanDisplayOptions()],
     (packages = [], displayOptions) => {
-      const { packageFilters, packageSorters } = displayOptions;
+      const { searchText, packageFilters, packageSorters } = displayOptions;
 
-      const filteredPackages = packages.filter((pkg) =>
+      const searchValue = searchText.toLowerCase();
+
+      // TODO: looks excessive with Array.isArray, but eslint will get trigger for 'use nullish coalescing' otherwise
+      //       with just ||, probably could be simplified
```
ctizen, line 55: Probably you could suppress the errors by adding `!!` to each part of OR-clause, so then you won't need that check

ctizen (changes requested):

packages/web/src/components/layouts/SearchResults/SearchResults.tsx
```diff
@@ -92,6 +115,21 @@ export default function SearchResults({
     },
   ];
 
+  const handleSortChange = useCallback(
+    (newSorterName: PackageSortType) => {
+      const sortOrder = newSorterName === sortField && sortDirection === 'DESC' ? 'ASC' : 'DESC';
```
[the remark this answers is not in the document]
KnoWabLe: If user clicks on sort button that is already active (by severity, for example) it will change direction to a opposite. If user clicks on a sort that is not active, it will active it and set direction to `DESC`

If you would left just the `sortDirection === 'DESC' ? 'ASC' : 'DESC'` part clicking on a new sort method, it will change sort method and order simultaneously, ending up in confusing UI like that:
19ef2d8d35

packages/web/src/components/ui/SidebarCategory/SidebarCategorySearch.tsx
KnoWabLe: To show user input while typing as-is and to update redux state via debounced callback after user is done typing to cut unnecessary and expensive sort/filter recalculations on every keystroke.

packages/web/src/components/pages/WebsiteResults.tsx
```diff
@@ -87,9 +111,47 @@ export function WebsiteResultsPage() {
...
   );
 
+  const selectedSortField = selectedDisplayOptions.packageSorters[0].by;
+  const selectedSortDirection = selectedDisplayOptions.packageSorters[0].direction;
+
+  const handleSortChange = useCallback(
+    (newSorterName: PackageSortType) => {
+      const sortOrder =
+        newSorterName === selectedSortField && selectedSortDirection === 'DESC' ? 'ASC' : 'DESC';
```
zardak, line 142: What are your opinion on moving this logic into the sorter component itself? So it receives current sorting order as a whole `PackageSorter` and passes back a new one in a cb?

packages/web/src/components/ui/SidebarCategory/SidebarCategorySearch.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  useEffect(() => {
+    setSearchValue(searchText);
+  }, [searchText]);
+
+  const searchChangeHandler = (inputValue: string) => {
```
zardak, line 26: This functions follows `handleSearchTextChange` interface in exact manner, why would we need it here?

packages/web/src/components/ui/SidebarCategory/SidebarCategorySearch.tsx
zardak: Seems like something that should be handled higher in component tree, rather than at a specialized input level.

packages/web/src/store/selectors/scanDisplayOptions/packageSorters.ts
```diff
@@ -15,6 +15,14 @@ export const compareSeverities = (severityA = 'UNKNOWN', severityB = 'UNKNOWN')
 type PackageSorterComparator = (pkgA: IdentifiedPackage, pkgB: IdentifiedPackage) => number;
 
 const packageSorterMap: Record<PackageSortType, PackageSorterComparator> = {
+  name: (pkgA, pkgB) => pkgB.name.localeCompare(pkgA.name),
+  size: (pkgA, pkgB) => {
+    if (pkgA?.approximateByteSize && pkgB?.approximateByteSize) {
```
zardak, line 20: We still want to sort things in cases like `pkA` has size, but `pkgB` doesn't. I'd suggest rewriting this as
`size: (pkgA, pkgB) => (pkgA?.approximateByteSize ?? 0) - (pkgB?.approximateByteSize ?? 0)`

packages/web/src/store/selectors/scanDisplayOptions/scanDisplayOptions.ts
```diff
@@ -46,14 +47,27 @@ export const makeSelectSortedAndFilteredScanPackages = () =>
...
     (packages = [], displayOptions) => {
-      const { packageFilters, packageSorters } = displayOptions;
+      const { searchText, packageFilters, packageSorters } = displayOptions;
 
-      const filteredPackages = packages.filter((pkg) =>
+      const searchValue = searchText.toLowerCase();
+
+      const matchedPackages = packages.filter(({ name, registryMetadata }) => {
+        return (
+          name.toLowerCase().includes(searchValue) ||
```
zardak, line 56: This could probably be simplified as follows
```
[name, registryMetadata?.description?, ...].some(field => field?.toLowerCase().includes(searchValue))
```

zardak (changes requested):

packages/web/src/store/selectors/scanDisplayOptions/scanDisplayOptions.ts
KnoWabLe: Just spreading, for example, `keywords` like this:
```js
[
  name,
  registryMetadata?.description,
  ...registryMetadata?.keywords
].some((field) => field?.toLowerCase().includes(searchValue));
```

Gives a error: `ESLint: Unsafe usage of optional chaining. If it short-circuits with 'undefined' the evaluation will throw TypeError`

So I came up with solution like this:
```js
const keywords = registryMetadata?.keywords?.length ? registryMetadata?.keywords : [];
const maintainers = registryMetadata?.maintainers?.length
  ? registryMetadata?.maintainers.map((it) => it.name)
  : [];

return [name, registryMetadata?.description, ...keywords, ...maintainers].some((field) =>
  field?.toLowerCase().includes(searchValue)
);
```

Is it better?

packages/web/src/components/pages/WebsiteResults.tsx
KnoWabLe: I can see that if sorter would be just one component. In our case we got two - desktop and mobile one. Would duplication of the handler locally on both components be worth it in this case?

packages/web/src/components/ui/SidebarCategorySearch/SidebarCategoryHiddenSearch.tsx
```diff
@@ -43,13 +44,13 @@ type Props = {
...
   clearInput: () => void;
-  alphabeticalGroups: GroupItem[];
+  alphabeticalGroups?: GroupItem[];
   selectedItems: string[];
   itemsWithImage?: boolean;
   selectHandler: (selectedKeywords: string[]) => void;
 };
 
-export default function SidebarCategorySearch({
+export default function SidebarCategoryHiddenSearch({
```
ctizen, line 53: Would you mind renaming scss file and folder as well?

ctizen (approved):

packages/web/src/components/pages/WebsiteResults.tsx
```diff
@@ -56,13 +58,16 @@ export function WebsiteResultsPage() {
 
   const availableFilters: PackageFilters = useMemo(
     () => ({
+      searchText: '',
```
zardak, line 61: We could possible mark this as an optional field in the type itself and avoid specifying it here.

packages/web/src/components/pages/WebsiteResults.tsx
```diff
@@ -56,13 +58,16 @@ export function WebsiteResultsPage() {
...
       authors: searchableEntities.packageAuthors.map((it) => it.name),
       keywords: searchableEntities.packageKeywords,
       traits: ['vulnerable', 'outdated'],
     }),
     [searchableEntities]
   );
 
+  const availableSorters: PackageSortType[] = ['name', 'size', 'popularity', 'severity'];
```
zardak, line 69: This should either be memo'd or moved to module-level const.

packages/web/src/components/pages/WebsiteResults.tsx
```diff
@@ -176,8 +226,12 @@ export function WebsiteResultsPage() {
         bundleSize={scanOverview.bundleSize ?? 0}
         scanDate={scanResult?.scan?.finishedAt}
         selectedFilters={selectedDisplayOptions.packageFilters}
+        selectedSort={selectedSort}
```
zardak, line 229: Why not keep the naming in line with other props, i.e. `selectedSorters`?

packages/web/src/components/pages/WebsiteResults.tsx
```diff
@@ -87,9 +92,54 @@ export function WebsiteResultsPage() {
...
+            ...selectedDisplayOptions,
+            packageSorters: newSorters,
+          },
+        })
+      );
+    },
+    [dispatch, normalizedUrl, selectedDisplayOptions]
+  );
+
+  const selectedSort = {
```
zardak, line 117: This could be simplified as `const selectedSort = selectedDisplayOptions.packageSorters[0];`

packages/web/src/components/pages/WebsiteResults.tsx
zardak: Valid point, LGTM for now.

packages/web/src/store/selectors/scanDisplayOptions/scanDisplayOptions.ts
zardak: Yep, this seems streamlined enough, LGTM

zardak (approved):

PR #113, opened 2022-10-18
feat: persist and serve page favicons

zardak:


packages/public-api/src/website/service.ts
```diff
@@ -108,10 +112,13 @@ export async function syncWebPageScanResult(scanReport: systemApi.ScanReport) {
 
     await webPageScanRepo.save(scanEntity);
 
-    if (scanEntity.status === 'processed') {
+    if (scanEntity.status === 'processed' && scanReport.status === 'ready') {
       await Promise.all([
         syncPackageUsageByHostname(scanEntity, em),
         syncScansWithVulnerabilities(scanEntity, em),
+        saveScanWebPageFavicon(scanReportUrl, scanReport.sourcePageMetadata).catch((e) =>
```
spalt08, line 119: Let's discuss on the next daily meeting if we upload icon inside `worker` vs `public-api`

spalt08 (approved):

ctizen (approved):

packages/public-api/src/website/service.ts
zardak: Absolutely valid point, but solving this in correct and stable matter would require a formidable amount of effort IMO. I'd better postpone it until we get to streaming scan results to front-end.

PR #117, opened 2022-10-24
feat: add more favicons

spalt08:
- Minor `server.ts` refactoring
- Added default fallback website favicon
- Added `favicon.ico`
- Changed sharing image


packages/web/src/components/Layout.tsx
```diff
@@ -40,6 +40,7 @@ export function Layout({
         <meta charSet='utf-8' />
         <meta name='viewport' content='width=device-width, initial-scale=1' />
         <meta name='mobile-web-app-capable' content='yes' />
+        <link rel='icon' type='image/x-icon' href='/static/favicon.ico' />
         {/* TODO: preload all fonts and add icons after redesign
           <link rel="preload" href={} as="font" type="font/woff2"/>
           <link rel='icon' type='image/svg+xml' href={'TODO'} />
```
ctizen, line 46: Looks like it's time to do something with this TODO

spalt08: I've created a ticket for that

ctizen (approved):

packages/web/src/server.tsx
```diff
@@ -13,41 +13,110 @@ import {
...
+        }
+
+        if (asset.endsWith('.css')) {
+          acc.css.push(path.join(publicStaticDir, asset));
+        }
+
+        return acc;
+      }, assets);
+  } catch (e) {
+    return assets;
```
zardak, line 110: nit: Seems like we'd rather throw here than return empty result?

packages/web/src/server.tsx
```diff
@@ -13,41 +13,110 @@ import {
...
+  }
+}
+
+async function pipeS3Object(objectKey: string, res: Response, fallback?: Buffer) {
+  try {
+    if (isDevelopment()) {
+      throw new Error('Development mode');
```
zardak, line 117: I'm rather unsure about this. Why not just let errors happen naturally?

packages/web/src/server.tsx
```diff
@@ -66,7 +135,9 @@ async function pipeS3Object(objectKey: string, res: Response) {
       throw new Error('Unexpected response format');
     }
   } catch (e) {
-    if (e instanceof NoSuchKey) {
+    if (fallback) {
+      res.send(fallback);
```
zardak, line 139: As we discussed, placing the fallback client-side would save us and users minor amount of traffic. And, more so, be a step towards correct implementation.

zardak (approved):

PR #119, opened 2022-10-25
feat: minor dev env changes

zardak:


cli/localSqsPuller.ts
```diff
@@ -13,12 +13,16 @@ import fetch from 'node-fetch';
  */
 
 // This hardcode should be in sync with values in local_start.sh
```
ctizen, line 15: Looks like the comment should be deleted as well

ctizen (approved):

cli/localSqsPuller.ts
zardak: `sqsQueueUrl` is still hardcoded, I'll get around it next time we touch configuration-related code

spalt08 (approved):

PR #121, opened 2022-10-26
fix: minor api fixup

zardak:


spalt08 (approved):

packages/public-api/src/middleware/common.ts
```diff
@@ -11,7 +11,7 @@ export const cors = createCorsMiddleware({
     if (originAllowList.length === 0 || (origin && originAllowList.includes(origin))) {
       callback(null, true);
     } else {
-      callback(new Error('Not allowed by CORS'));
+      callback(null, false);
```
spalt08, line 14: Are you sure it will work?

From docs:
image

zardak: Yes, I'm fairly sure. Per said docs: 
```
Function - set origin to a function implementing some custom logic. The function takes the request origin as the first parameter and a callback (called as callback(err, origin), where origin is a non-function value of the origin option) as the second.
```

ctizen (approved):

PR #125, opened 2022-10-27
fix: remove redundant fields

ctizen:


packages/public-api/src/packageMetadata/packageMetadataService.ts
```diff
@@ -1,23 +1,32 @@
...
     .createQueryBuilder('pm')
+    .select(columnsToSelect.map((col) => col[0]))
     .where('pm.name IN (:...packageNames)', { packageNames })
-    .getMany();
-
-  const packageMetadataMap = packageMetadataEntities.reduce((acc, val) => {
-    acc[val.name] = val;
+    .getRawMany();
 
+  return packageMetadataEntities.reduce((acc, val) => {
```
spalt08, line 24:
```js
 const packageMetadataMap = packageMetadataEntities.reduce((acc, val) => {
    // comment why
    delete val.fullDescription;
    delete val.versionSpecificValues;
    acc[val.name] = val;
}
```

packages/public-api/src/packageMetadata/packageMetadataService.ts
```diff
@@ -1,23 +1,32 @@
 import { getDatabaseConnection, PackageMetadata } from '@gradejs-public/shared';
 
-export async function getPackageMetadataByPackageNames(packageNames: string[]) {
+type Metadata = Omit<Omit<PackageMetadata, 'fullDescription'>, 'versionSpecificValues'>;
+
+export async function getPackageMetadataByPackageNames(
```
spalt08, line 5: Can we rename it to something like `getPackagePartialMetadataByPackageNames`?

packages/public-api/src/packageMetadata/packageMetadataService.ts
ctizen: Used the `select: false` solution eventually.

spalt08 (approved):

zardak (approved):

PR #126, opened 2022-10-27
fix: log db init errors

ctizen:


spalt08 (commented):
This may be bad since the app will be deployed still. Can we throw an error inside catch block?

spalt08 (approved):

PR #129, opened 2022-10-28
fix: favicon proper types

ctizen:


spalt08 (approved):

packages/public-api/src/website/metadata/favicon.ts
```diff
@@ -12,9 +15,33 @@ export async function saveScanWebPageFavicon(
...
+      }, FAVICON_TIMEOUT);
+    }),
+  ]);
+
+  if (!faviconRequest) {
+    return;
+  }
 
   const faviconContentType = faviconRequest.headers.get('content-type') ?? undefined;
+  if (faviconContentType && !VALID_MIME_TYPES.includes(faviconContentType)) {
```
bayotop, line 41: Might be better to be strict here and return if the `content-type` header is missing.

spalt08:
@zardak @ctizen 

There are two issues, one is someone using our S3 buckets as malware storage or something similar, and the other is attacking our infrastructure through SSRF.

The content-type check does help with the latter because someone external can't influence our responses. But the first one is still possible by forging the content type.

Ideally we should create a separate origin (like `gradejsusercontent.com` for serving untrusted static files)

PR #135, opened 2022-11-04
feat: search suggestions

KnoWabLe:


packages/web/index.html
```diff
@@ -31,5 +31,6 @@
 
     <div id="app"></div>
     <div id="modal-root"></div>
+    <div id="search-root"></div>
```
spalt08: This won't work since `index.html` is only used by vercel deploy previews. Consider modifying `components/Layout.tsx` for the production SSR build

PR #136, opened 2022-11-04
feat: initial search api implementation

zardak:


packages/shared/src/database/migrations/1667551928553-FulltextSearchIndexes.ts
```diff
@@ -0,0 +1,19 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class FulltextSearchIndexes1667551928553 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+        CREATE EXTENSION IF NOT EXISTS pg_trgm;
+        
+        CREATE INDEX "hostname_hostname_trgm" ON "hostname" USING gin ("hostname" gin_trgm_ops);
+        CREATE INDEX "package_popularity_view_package_name_trgm" ON "package_popularity_view" USING gin ("package_name" gin_trgm_ops);
```
spalt08, line 9: maybe on `package_metadata` with `monthly_downloads` as rank? So we can also select `description`

packages/public-api/src/search/searchService.ts
```diff
@@ -0,0 +1,44 @@
...
+    .createQueryBuilder('package')
+    .select('package.packageName')
+    .where('package.packageName like :namePartial', { namePartial: `%${filteredNamePartial}%` })
+    .orderBy('package.usageByHostnameCount', 'DESC')
+    .limit(SEARCH_RESULT_PER_ENTITY_LIMIT)
+    .getMany();
+
+  const hostnameSearchQuery = hostnameRepo
+    .createQueryBuilder('hostname')
+    .select('hostname.hostname')
```
spalt08: `.join(WebPage, "webpage", "webpage.hostname_id = hostname.id")` + join `WebPageScan` and `getRawMany`

zardak: This will return all of the scans for each hostname with no way to limit them. I'll add a proper subquery later in.

ctizen (approved):

spalt08 (approved):

